PR #10, opened 2020-03-22
add a contrib to build a gource log file

flepied:
Details about Gource at https://gource.io/


contrib/gource-project-log.py
```diff
@@ -0,0 +1,96 @@
...
+        print('%.0f|%s|A|/%s/%s|' % (created_time.timestamp(), entry['author'],
+                                     entry['repository_fullname'],
+                                     entry['title']))
+        print('%.0f|%s|M|/%s/%s|' % (merge_time.timestamp(), entry['author'],
+                                     entry['repository_fullname'],
+                                     entry['title']))
+        title[entry['repository_fullname_and_number']] = entry['title']
+
+    data = db.run_named_query(
+        'all_commented_changes',
```
morucci: I think that this new query is not needed. Instead use _scan and set param['etype'].

flepied: ok done

PR #11, opened 2020-03-23
changes_review.js: display duration in days, hours, minutes and secs

flepied:


web/src/components/changes_review.js
```diff
@@ -15,6 +15,20 @@ import {
 
 import { Line } from 'react-chartjs-2';
 
+function secondsToDhms(seconds) {
```
morucci: Great, just tried an it work well ! But I think we should use an existing lib for that. Could you try https://momentjs.com/docs/#/durations/ ? There is also an humanize function that seems nice https://momentjs.com/docs/#/durations/humanize/

flepied: ok I'll try that

flepied: done

morucci:
Nice thanks !

morucci:
Sorry I've clock the wrong button and close the PR instead of merging it :(. It is merged now +1

PR #12, opened 2020-03-23
improve UI

flepied:
- link to github PR/gerrit changes and author

### DIFF
--- a/monocle/gerrit/review.py
+++ b/monocle/gerrit/review.py
@@ -40,8 +40,8 @@ class ReviewesFetcher(object):
 
     log = logging.getLogger("monocle.ReviewesFetcher")
 
-    def __init__(self, host, repository_prefix):
-        self.host = host
+    def __init__(self, base_url, repository_prefix):
+        self.base_url = base_url
         self.repository_prefix = repository_prefix
         self.status_map = {
             'NEW': 'OPEN',
@@ -83,7 +83,7 @@ class ReviewesFetcher(object):
         reviews = []
         while True:
             urlpath = (
-                self.host + '/changes/' + request_params +
+                self.base_url + '/changes/' + request_params +
                 '&n=%s&start=%s' % (
                     count, start_after))
             self.log.info("query: %s" % urlpath)
@@ -132,6 +132,7 @@ class ReviewesFetcher(object):
                 'repository_fullname': review['project'],
                 'repository_shortname': "/".join(
                     review['project'].split('/')[1:]),
+                'url': '%s/%s' % (self.base_url, review['_number']),
                 'author': "%s/%s" % (
                     review['owner'].get('name'),
                     review['owner']['_account_id']),
--- a/monocle/github/pullrequest.py
+++ b/monocle/github/pullrequest.py
@@ -50,10 +50,10 @@ class PRsFetcher(object):
 
     log = logging.getLogger("monocle.PRsFetcher")
 
-    def __init__(self, gql, host, org, bulk_size=25):
+    def __init__(self, gql, base_url, org, bulk_size=25):
         self.gql = gql
         self.size = bulk_size
-        self.host = host
+        self.base_url = base_url
         self.org = org
         self.events_map = {
             'ClosedEvent': 'ChangeAbandonedEvent',
@@ -241,6 +241,7 @@ class PRsFetcher(object):
                 'number': change['number'],
                 'repository_fullname_and_number': change[
                     'repository_fullname_and_number'],
+                'url': change['url'],
                 'on_author': change['author'],
                 'on_created_at': change['created_at'],
             })
@@ -259,6 +260,10 @@ class PRsFetcher(object):
                 change['repository_fullname'],
                 change['number'],
             )
+            change['url'] = ('%s/%s/pull/%s'
+                             % (self.base_url,
+                                change['repository_fullname'],
+                                change['number']))
             change['author'] = pr['author']['login']
             change['branch'] = pr['headRefName']
             change['target_branch'] = pr['baseRefName']
--- a/web/src/components/common.js
+++ b/web/src/components/common.js
@@ -5,6 +5,25 @@ import Col from 'react-bootstrap/Col'
 import Card from 'react-bootstrap/Card'
 
 
+function change_url(x, name=null) {
+  if (!name) {
+    name = x.repository_fullname_and_number;
+  }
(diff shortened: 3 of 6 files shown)


flepied:
@morucci I don't know how to do it on the "Events Stats" tab

web/src/components/common.js
```diff
@@ -5,6 +5,27 @@ import Col from 'react-bootstrap/Col'
 import Card from 'react-bootstrap/Card'
 
 
+function change_url(x, name=null) {
```
morucci: We should probably add the review system base url in each Change and Change events in the DB. And makes the web UI able to react on that. We should be able to handle any gerrit instances as well and also any github enterprise deployments.

web/src/components/top.js
```diff
@@ -34,7 +34,7 @@ class TopEventsTable extends React.Component {
                 <tbody>
                   {this.props.data.map((x, index) =>
                     <tr key={index}>
-                      <td>{index}</td>
+                      <td>{index + 1}</td>
```
morucci: This change could be done in another PR imo.

web/src/components/common.js
flepied: sure it would be cleaner

web/src/components/top.js
flepied: ok

web/src/components/common.js
flepied: @morucci where should I do that in the code?

web/src/components/top.js
flepied: done

web/src/components/common.js
flepied: I tried to do it for github pr. let me know if it was the right way

monocle/github/pullrequest.py
```diff
@@ -257,6 +258,10 @@ def extract_pr_objects(pr):
                 change['repository_fullname'],
                 change['number'],
             )
+            change['url'] = ('https://%s/%s/pull/%s'
```
morucci: I would have instead added three fields:
- host: could be github.com or github.company.com or gerrit.ovirt.com or softwarefactory-project.io
- host_type: could be gerrit or github
- host_base_url: should be None usually for most services accessible from '/' but must be '/r' for softwarefactory-project.io for instance.

From those fields, in the UI, we can compute the PR web url.

This three fields must be added in db.py self.index. A index deletion + re-index will be needed. And host_base_url it should be an option CLI argument in the crawler UI in bot gerrit and github crawler I guess.

Let me know if you think about a better way to achieve this.

flepied: I think I'll prefer to have the url computed on the backend side to have the simplest API. Let's discuss when you'll have time.

morucci: Ok fine to me, the change url could be computed in the in the crawler and be part of object in the DB. What about the url to access user profils ? Should it be computed in the same way ? I'd say yes if we move this way. We should also take care of the host_base_url.

flepied: user profiles should be done the same way I'd say too. Should we just pass a base_url  to the crawlers?

Longer term we could also have internal pages for users and changes.

morucci: Yes. And looking more at the code. I see we could use the existing --host option. For the gerrit crawler it is well handled, I can call the gerrit crawler this way:
```
python3 ./monocle/main.py gerrit_crawler --repository "^software-factory/.*" --updated-since "2019-01-01 00:00:00" --loop-delay 60 --host https://softwarefactory-project.io/r
```

But for github, it is not consistent because it does not include the scheme:
```
monocle github_crawler --host github.com --token XXX --org docker --updated-since 2020-03-01
```
So it should be a complete url with the scheme, and looking at graphql.py self.url is hardcoded. So this should be fixed as well.

So yes we could use --host directly.

monocle/gerrit/review.py
```diff
@@ -40,8 +40,8 @@ class ReviewesFetcher(object):
 
     log = logging.getLogger("monocle.ReviewesFetcher")
 
-    def __init__(self, host, repository_prefix):
-        self.host = host
+    def __init__(self, base_url, repository_prefix):
+        self.base_url = base_url
```
morucci: This could be renamed to base_url in the CLI as well.

monocle/github/pullrequest.py
```diff
@@ -257,6 +258,10 @@ def extract_pr_objects(pr):
                 change['repository_fullname'],
                 change['number'],
             )
+            change['url'] = ('%s/%s/pull/%s'
```
morucci: You need to add "url" in elastic schema in db.py.

morucci:
Also could it be splitted in two PRs ?:
* backend change
* ui change

monocle/gerrit/review.py
```diff
@@ -130,6 +130,7 @@ def extract_pr_objects(review):
                 'repository_fullname': review['project'],
                 'repository_shortname': "/".join(
                     review['project'].split('/')[1:]),
+                'url': '%s/c/gerrit/+/%s' % (self.base_url, review['_number']),
```
morucci: This works better:
'url': '%s/%s' % (self.base_url, review['_number']),

monocle/github/pullrequest.py
morucci: See this change: https://github.com/morucci/monocle/pull/28

monocle/gerrit/review.py
flepied: done

monocle/github/pullrequest.py
flepied: ok

monocle/gerrit/review.py
flepied: done

monocle/github/pullrequest.py
flepied: is it ok now?

web/src/components/common.js
```diff
@@ -5,6 +5,25 @@ import Col from 'react-bootstrap/Col'
...
+  if (!name) {
```
morucci: Could you move the JS part in another PR ?

flepied: I split it in 2 commits

morucci:
I've tested it. It works with a quick change in the common.js. Expect the author_url, that might require a bit more work. So I merge the PR. I'll followup with a quick fix.

morucci:
See #31

PR #13, opened 2020-03-23
Avoid dirty stacktrace when query match no data

morucci:


morucci:
Hi would you mind to review @flepied ?

PR #24, opened 2020-03-25
store the PR description

flepied:


morucci:
Do we want to enable full text search on that field ?
Also I wanted to check if with gerrit we could get the same data (the review commit message) and it work as well: https://github.com/morucci/monocle/pull/29

flepied:
> Do we want to enable full text search on that field ?

Yes I think it could be good for future work to have it indexed full text.

morucci:
see #30  that add the text field.

Issue #4, opened 2019-12-24
Add the branch in the Change object


morucci:

flepied:
that could help in some analysis I would like to do like being sure a ticker is linked to the PR. Some ticketing systems allow to have the ticket number in the branch name, in the title or in the description.

morucci:
Ok in that case you mean the "from" branch. Have a ticket number in the "from" branch. Here it about the PR target branch but both are nice piece of information. As the "from" branch make sense in the PR model, it does not in a Gerrit/Review model. But it can be set to "master" by default for gerrit maybe ?

flepied:
yes I agree we need both branches

PR #26, opened 2020-03-25
Add branch and target_branch in db schema

morucci:
related to issue #4


flepied:
do we need to add text also for the change message?

morucci:
@flepied yes, I'll do it in another PR. I'm merging this one.

morucci:
@flepied see #30  that add the text field.

PR #35, opened 2020-03-29
Add docker-compose support

flepied:
Issue: https://github.com/morucci/monocle/issues/19


monocle/db/db.py
```diff
@@ -19,20 +19,33 @@
...
-
 
 class ELmonocleDB():
 
-    def __init__(self, index='monocle'):
-        self.es = client.Elasticsearch('localhost:9200')
+    def __init__(self, elastic_conn='localhost:9200', index='monocle'):
+        while True:
+            try:
+                print('Connecting to ES server at %s' % elastic_conn)
```
morucci: You could use a logger.

monocle/envdefault.py
```diff
@@ -0,0 +1,41 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+
+import argparse
```
morucci: Once the generic crawler is merged, we should not need this env vars input. The configuration should be passed through a config yaml file. And called with: monocle crawler --config config.yaml

morucci: But, this could be refactored after we merge both PRs.

monocle/webapp.py
```diff
@@ -30,14 +31,15 @@
...
+CORS(app, resources={r"/api/0/query/*":
+                     {"origins": os.getenv('ALLOW_ORIGIN', '*')}})
 
 
 @app.route("/api/0/query/<name>", methods=['GET'])
 def query(name):
     repository_fullname = request.args.get('repository')
     params = utils.set_params(request.args)
-    db = ELmonocleDB()
+    db = ELmonocleDB(os.getenv('ELASTIC_CONN', 'localhost:9200'))
```
morucci: As the api will be nested in the monocle cli, we could remove the env vars input as well.

morucci: But again can be fixed wth a followup pr.

README.md
```diff
@@ -14,60 +14,37 @@ cd monocle
 mkdir data
 ```
```
morucci: chmod o+w data

w/o it elasticsearch does not start. Forbidden to write in data.

monocle/envdefault.py
flepied: sure it depends which one we merge first...

README.md
flepied: ok changing that

flepied: but it works for me without it btw...

monocle/db/db.py
flepied: ok changing that

flepied: done

README.md
flepied: done

PR #39, opened 2020-03-31
implement a changes by repository box in the ui

flepied:


web/src/App.js
```diff
@@ -91,6 +92,12 @@ class RootView extends React.Component {
                   <CLastChanges />
                 </Col>
               </Row>
+              <Row>
+                <Col>
+                  <CRepoChanges />
+                </Col>
+              </Row>
+              <Row><Col><p></p></Col></Row>
```
morucci, line 100: can be removed

web/src/App.js
```diff
@@ -91,6 +92,12 @@ class RootView extends React.Component {
                   <CLastChanges />
                 </Col>
               </Row>
+              <Row>
+                <Col>
+                  <CRepoChanges />
```
morucci, line 97: I'm wondering to have a repository related events tab. Now we have Events stats, changes status, what about a Repo stats. Even if we will rework layout and where graph/stat should be, I think there is a room for that new tab:

- Top repository by merged changes
- Top repository by open changes
- Top 3 authors by repo
- Less active repositories
- ...

web/src/components/common.js
```diff
@@ -12,6 +12,24 @@ function change_url(x, name=null) {
   return <a href={x.url} target="_blank" rel="noopener noreferrer">{name}</a>;
 }
 
+function filter_url(url, field) {
```
morucci, line 15: You should use this: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get

web/src/components/common.js
```diff
@@ -12,6 +12,24 @@ function change_url(x, name=null) {
...
+  
+  if (url_parts.length > 1) {
+    const new_args = url_parts[1].split('&').filter(x => !x.startsWith(field + '='));
+    
+    return url_parts[0] + '?' + new_args.join('&');
+  }
+  return url;
+}
+
+function add_url_field(url, field) {
```
morucci, line 26: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/set

morucci: See also here, its already used: https://github.com/morucci/monocle/blob/master/web/src/components/filtersform.js#L77

monocle/db/queries.py
```diff
@@ -249,6 +249,12 @@ def _events_top(
         'count_avg': count_avg, 'count_median': count_median}
 
 
+def merged_changes_by_repository(es, index, repository_fullname, params):
```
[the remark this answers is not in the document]
flepied: ok done

web/src/App.js
flepied: yes good idea but in following PR

web/src/App.js
flepied: done

web/src/App.js
flepied: yes in a following PR

web/src/components/common.js
flepied: ok done

web/src/components/common.js
flepied: done

PR #41, opened 2020-04-01
cleaner start of crawler

flepied:
- add dependencies between services in docker-compose.yml
- test the connection before using the elastic API

### DIFF
--- a/monocle/db/db.py
+++ b/monocle/db/db.py
@@ -20,6 +20,7 @@
 # SOFTWARE.
 
 import logging
+import socket
 import time
 
 from datetime import datetime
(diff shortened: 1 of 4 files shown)


monocle/db/db.py
```diff
@@ -36,17 +37,26 @@ class ELmonocleDB():
...
-                self.log.info(self.es.info())
+                s.connect((ip, int(port)))
+                s.shutdown(2)
+                s.close()
                 break
-            except Exception:
-                self.log.exception('Unable to connect to %s. Sleeping.'
-                                   % elastic_conn)
+            except Exception as excpt:
+                self.log.info('Unable to connect to %s: %s. Sleeping.'
```
morucci, line 52: Why not using self.log.exception instead ?

monocle/db/db.py
```diff
@@ -36,17 +37,26 @@ class ELmonocleDB():
...
+                              % (elastic_conn, excpt))
                 time.sleep(10)
```
morucci, line 54: Could you set the timeout in a var

monocle/db/db.py
morucci, line 52: 'Sleeping for %s' % cnx_sleep_timeout

flepied: because it's too verbose

flepied: ok done

monocle/db/db.py
flepied: done

monocle/main.py
```diff
@@ -131,7 +134,8 @@ def main():
                     loop_delay=project['crawler']['loop_delay'],
                     token=crawler_item['token'],
                     base_url=crawler_item['base_url'])
-                tpool.append(Crawler(c_args, elastic_conn=args.elastic_conn))
+                tpool.append(Crawler(c_args, elastic_conn=args.elastic_conn,
+                                     elastic_timeout=args.elastic_timeout))
```
morucci, line 138: You need that change on the gerrit crawler too, few lines below.

monocle/db/db.py
```diff
@@ -35,17 +36,27 @@ class ELmonocleDB():
...
+                s.connect((ip, int(port)))
+                s.shutdown(2)
+                s.close()
                 break
-            except Exception:
-                self.log.exception('Unable to connect to %s. Sleeping.'
-                                   % elastic_conn)
-                time.sleep(10)
+            except Exception as excpt:
+                self.log.info('Unable to connect to %s: %s. Sleeping for %ds.'
```
morucci, line 53: Don't you want to use self.log.exception ?

flepied: nope because it was done on purpose to avoid to scare people for nothing

monocle/main.py
flepied: done

PR #49, opened 2020-04-04
Authors exclude

morucci:


morucci:
I'd like to handle that case in all entry point, cli, api, ui.
The UI disable the exclude-authors filed when a authors filter is set, but also does not call the api with the exclude-authors param.

PR #56, opened 2020-04-07
webui: set viewport container as fluid to get width: 100%

morucci:


flepied:
being used to the github ui I think I prefer not to merge this change because we don't have a lot of info to put horizontally and in the future we could add a column on the right or on the left for other usage like a menu like on github.

PR #68, opened 2020-04-10
web: ChangesTable add column with Complexity about changes

flepied:
Complexity = (files+additions+deletions)

### DIFF
--- a/web/src/components/changes.js
+++ b/web/src/components/changes.js
@@ -10,6 +10,8 @@ import Table from 'react-bootstrap/Table'
 import ReactPaginate from 'react-paginate'
 import PropTypes from 'prop-types'
 
+import moment from 'moment'
+
 import {
   BaseQueryComponent,
   LoadingBox,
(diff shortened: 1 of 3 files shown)


web/src/components/changes.js
```diff
@@ -137,6 +138,7 @@ class ChangesTable extends React.Component {
                       <td><a href={addUrlField('authors', x.author)}>{x.author}</a></td>
                       <td>{changeUrl(x, x.title)}</td>
                       {this.props.mergeable ? <td>{x.mergeable}</td> : null}
+                      <td align="center">{ x.changed_files + x.additions + x.deletions}</td>
```
morucci, line 141: Could you extract into a function ?

Also what about a complexity function such as:
m = additions + deletions
score_m = 1 if m < 10
socre_m = 2 if m < 100
...
fc = file changes count
score_fc =  1 if fc < 2
score_fc = 2 if fc < 4
...

score = score_m + score_fc + ...

?

flepied: I have factorized this code in the reducers to make it more useful globally. Btw I have added a bubble graph in the /merged-changes route

PR #80, opened 2020-04-13
Improve UX

flepied:
- web: add a timeline to the ChangeTable
- web: do not display the time if not needed 
- web: add a /abandoned-changes/ page
- web: add an Abandoned Changes table to the main view
- abandoned_changes returns CLOSED Changes
- fix the number of elements returned for hot_changes and cold_changes
- list the total number of changes in opened-changes and merged-changes
- complexity graph in merged-changes use duration instead of time and has log X scale
- tables of merge changes now has a duration column

### DIFF
--- a/monocle/db/queries.py
+++ b/monocle/db/queries.py
@@ -430,6 +430,7 @@ def first_review_on_changes(es, index, repository_fullname, params):
 
 def cold_changes(es, index, repository_fullname, params):
     params = deepcopy(params)
+    size = params.get('size')
     params['etype'] = ('Change',)
     params['state'] = 'OPEN'
     changes = _scan(es, index, repository_fullname, params)
@@ -442,15 +443,15 @@ def cold_changes(es, index, repository_fullname, params):
     changes_wo_rc = [
         change for change in changes if change['change_id'] in changes_ids_wo_rc
     ]
-    return {
-        'items': sorted(
-            changes_wo_rc, key=lambda x: dbdate_to_datetime(x['created_at'])
-        )
-    }
+    items = sorted(changes_wo_rc, key=lambda x: dbdate_to_datetime(x['created_at']))
+    if size:
+        items = items[:size]
+    return {'items': items}
 
 
 def hot_changes(es, index, repository_fullname, params):
     params = deepcopy(params)
+    size = params.get('size')
     # Set a significant depth to get an 'accurate' median value
     params['size'] = 500
     top_commented_changes = changes_top_commented(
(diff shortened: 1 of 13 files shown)


morucci:
I can't access 'http://localhost:3000/monocle/abandoned-changes/' this is a blank page.

monocle/db/queries.py
```diff
@@ -641,3 +645,18 @@ def changes_and_events(es, index, repository_fullname, params):
     data = run_query(es, index, body)
     changes = [r['_source'] for r in data['hits']['hits']]
     return {'items': changes, 'total': data['hits']['total']}
+
+
+def abandoned_changes(es, index, repository_fullname, params):
```
morucci, line 650: You could keep the same naming rule: last_abandoned_changes ?

web/src/components/change.js
```diff
@@ -30,6 +30,8 @@ class ChangeStatus extends React.Component {
         return <Badge variant="success">Open</Badge>
       case 'MERGED':
         return <Badge variant="primary">Merged</Badge>
+      case 'CLOSED':
+        return <Badge variant="danger">Closed</Badge>
```
morucci, line 35: Why not using the Abandoned term ?

web/src/components/common.js
```diff
@@ -36,6 +36,17 @@ function addS (count, s = 's') {
   }
 }
 
+function prepareLabels (histo) {
```
morucci, line 39: What about fixing this in how elasticsearch returns the date: https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-aggregations-bucket-datehistogram-aggregation.html#_keys_2

web/package.json
```diff
@@ -12,6 +12,7 @@
     "react-chartjs-2": "^2.9.0",
     "react-datepicker": "^2.14.1",
     "react-dom": "^16.13.1",
+    "react-event-timeline": "^1.6.3",
```
morucci, line 15: Nice !

web/src/components/change.js
flepied: do we want to change CLOSED to ABANDONED too?

web/src/components/common.js
flepied: done

monocle/db/queries.py
flepied: done

web/src/components/change.js
morucci: No, let's keep like that. It is called at various places closed in the backed.

PR #86, opened 2020-04-15
github crawler: skip archived repositories

### LINKED ISSUE #57: archived repositories
Information from archived repositories should not be fetched and displayed.

morucci:
Fix #57

### DIFF
--- a/monocle/github/pullrequest.py
+++ b/monocle/github/pullrequest.py
@@ -177,7 +177,9 @@ class PRsFetcher(object):
             scope = "repo:%(org)s/%(repository)s" % kwargs
         kwargs['scope'] = scope
         qdata = '''{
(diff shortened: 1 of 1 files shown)


flepied:
What will happen if the repo is archived with some open PR we have already stored?

morucci:
Those PRs will still be in the DB and appear in stats. In the DB currently we don't have indicators from which kind of project come this PR.
At that point you need to wipe and re-index. Or dbmanage delete the archived repositories.

flepied:
could we do it another way like to get the archive status instead of filtering on it?

morucci:
We can from the graphql query get the PR's related repository and get the archive attribute. Then we could add that setting as a supported query params.
But, and that's not solve as well with the current proposed implementation what to do when a repository is set as archived so we have previously fetched PRs with repo_archived=False and new PRs with repo_archived=True ?

morucci:
@flepied I see that PRs on a project I passed to archived are not updated. 'updated_at' field does not change. So the approach we discussed won't work I assume.

PR #92, opened 2020-04-16
added pie charts for opened-changes and merged-changes

flepied:


web/src/components/approvals.js
```diff
@@ -83,8 +45,9 @@ class ApprovalStats extends BaseQueryComponent {
               <Card.Body>
                 <Row>
                   <Col>
-                    <ChangeApprovals
-                      data={data}
+                    <Pie
```
morucci, line 48: Could we override the palette for the approval ? I'd like to keep the red colors for negative approval and green for +1, +2, APPROVED.

flepied: done

morucci:
Since the queries test PR merged we have a list of public queries [1]. The two new queries must be added in the list.

[1] https://github.com/morucci/monocle/blob/master/monocle/db/queries.py#L35

PR #94, opened 2020-04-17
web: add global menu entries

flepied:


web/src/components/menu.js
```diff
@@ -1,13 +1,22 @@
...
   render () {
+    const search = window.location.search
     return <Navbar bg="light" expand="lg">
-      <Navbar.Brand ><Link className="navbar-brand" to={`/r/${this.props.index}`}>Monocle</Link></Navbar.Brand>
+      <Navbar.Brand>
+        <Link className="navbar-brand" to={`/r/${this.props.index}`}>Monocle</Link>
+      </Navbar.Brand>
+      <Nav className="mr-auto">
+        <Link className="nav-link" to={`/${this.props.index}${search}`}>Top</Link>
+        <Link className="nav-link" to={`/${this.props.index}/merged-changes${search}`}>Merged</Link>
```
morucci, line 17: What about a "dropdown menu" for "Changes" that contains "Merged Changes" and "Open Changes" (https://getbootstrap.com/docs/4.0/components/navbar/) ?

flepied: I think we'll switch to dropdown when we'll have more entries.

morucci: Ok so I merge :) !

PR #102, opened 2020-04-18
crawler: solidify github process

flepied:
- manage errors from github
- run a thread per Github token
- dump raw pr into the /var/lib/crawler if it exists
- manage empty authors as ghost
- retry multiple times if a graphql request fails

### DIFF
--- a/monocle/crawler.py
+++ b/monocle/crawler.py
@@ -100,18 +100,18 @@ class Runner(object):
             log.info("%d objects will be updated in the database" % len(objects))
             self.db.update(objects)
 
-    def run(self):
-        while True:
-            self.run_step()
-            log.info("Waiting %s seconds before next fetch ..." % (self.loop_delay))
-            sleep(self.loop_delay)
-
 
 class Crawler(Thread, Runner):
     def __init__(self, args, elastic_conn='localhost:9200', elastic_timeout=10):
         Runner.__init__(self, args, elastic_conn, elastic_timeout)
         Thread.__init__(self)
+
+    def run(self):
         self.setName(self.repository_el_re)
+        while True:
+            self.run_step()
+            log.info("Waiting %s seconds before next fetch ..." % (self.loop_delay))
+            sleep(self.loop_delay)
 
 
 class GroupCrawler(Thread):
--- a/monocle/crawler.py
+++ b/monocle/crawler.py
@@ -19,8 +19,10 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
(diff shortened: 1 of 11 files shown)


monocle/crawler.py
```diff
@@ -53,31 +58,90 @@ def __init__(self, args, elastic_conn='localhost:9200', elastic_timeout=10):
...
+                    )
+                    self.db.update(objects)
+                break
+            except pullrequest.ExtractPRIssue as err:
+                log.exception('Unable to decode PR')
+                log.error('idx=%d pr=%s' % (err.idx, err.pr))
+                # stop if we were on the last pr
+                if len(prs) == err.idx + 1:
+                    break
+                prs = prs[: err.idx] + prs[err.idx + 1 :]  # noqa
```
morucci, line 116: I don't get why we need to dump everything line 99.
Why to process extract_objects prs by bulk. It seems easier to me to skip and dump a failed extract in each gerrit and github crawler code directly. There is already the code to skip in each crawler so we can pass a dump_on_failed function to extract_objects.

monocle/github/pullrequest.py
```diff
@@ -258,6 +268,11 @@ def get_one(self, org, repository, number):
         return (raw, self.extract_objects([raw]))
 
     def extract_objects(self, prs):
+        def get_login(data):
+            if data and 'login' in data and data['login']:
+                return data['login']
+            return 'ghost'
```
morucci, line 274: It is fine to have author at None.

flepied: this is the github account that is used for deleted users

monocle/crawler.py
flepied: yes changing that

monocle/github/pullrequest.py
```diff
@@ -386,24 +405,44 @@ def extract_pr_objects(pr):
                     'created_at': _commit.get('pushedDate', change['created_at']),
                 }
                 if _commit['author'].get('user'):
-                    obj['author'] = _commit['author']['user'].get('login')
+                    obj['author'] = get_login(_commit['author']['user'])
                 insert_change_attributes(obj, change)
                 objects.append(obj)
             return objects
 
+        def dump_data(data, prefix=None):
```
morucci, line 413: Do you think that's possible to have the same dump support for the Gerrit crawler ?

flepied: sure but let's do that in another PR

morucci: Ok. So let's merge this PR. I'm adding an issue to not forget about adding the same support for the Gerrit crawler.

flepied: done

PR #126, opened 2020-04-22
manage the draft status from Gihub PR

flepied:


monocle/github/pullrequest.py
```diff
@@ -296,6 +297,7 @@ def extract_pr_objects(pr):
             change = {}
             change['type'] = 'Change'
             change['id'] = pr['id']
+            change['draft'] = pr['isDraft']
```
morucci, line 300: You need to add that new field in the schema.

morucci: Could you add an issue to add the same support in Gerrit ?

flepied: is the same feature available in gerrit?

morucci: Seems there was a draft support that has been removed. There is a "work in progress"  https://www.gerritcodereview.com/2.15.html#support-for-draft-changes-removed
Well let's manage that if when someone will require it.

morucci: In https://github.com/morucci/monocle/blob/master/monocle/db/db.py#L101 add "draft": {"type": "boolean"},

PR #132, opened 2020-04-23
web: visualize changes on files using treemaps

flepied:
- new requests: changes_by_file_map and authors_by_file_map
- files param (regexp to filter on changed_files.path)

### DIFF
--- a/monocle/db/queries.py
+++ b/monocle/db/queries.py
@@ -779,10 +779,7 @@ def changes_by_file_map(es, index, repository_fullname, params):
     for change in changes:
         for f in change['changed_files']:
             key = '{}:{}'.format(change['repository_fullname'], f['path'])
-            try:
-                files[key] += 1
-            except KeyError:
-                files[key] = 1
+            files[key] = files.get(key, 0) + 1
     return {'changes': files}
 
 
--- a/monocle/db/queries.py
+++ b/monocle/db/queries.py
@@ -62,6 +62,8 @@ public_queries = (
     "changes_and_events",
     "last_abandoned_changes",
     "new_contributors",
+    "changes_by_file_map",
+    "authors_by_file_map",
 )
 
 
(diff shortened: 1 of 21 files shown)


monocle/db/queries.py
```diff
@@ -759,3 +764,44 @@ def new_contributors(es, index, repository_fullname, params):
     )
     diff = new.difference(old)
     return {'items': [n for n in new_authors if n['key'] in diff]}
+
+
+def changes_by_file_map(es, index, repository_fullname, params):
```
morucci, line 769: Could you add the support for size. And also order the returned values.

monocle/db/queries.py
```diff
@@ -759,3 +764,44 @@ def new_contributors(es, index, repository_fullname, params):
...
+        for f in change['changed_files']:
+            key = '{}:{}'.format(change['repository_fullname'], f['path'])
+            try:
+                files[key] += 1
+            except KeyError:
+                files[key] = 1
+    return {'changes': files}
+
+
+def authors_by_file_map(es, index, repository_fullname, params):
```
morucci, line 789: Do you think we could get the change rate by file for each authors in the returned values ?

monocle/db/queries.py
flepied: the values are in a dict so there is no need for sorting.

monocle/db/queries.py
flepied: we'll have to define what it really means but I'd prefer to that in another PR.

morucci:
I think that's a nice to have this treemap. I was able to navigate and saw my changes in zuul/.* The tooltip is hard to understand "1MB, 7/9" <- What is this ?

flepied:
> I think that's a nice to have this treemap. I was able to navigate and saw my changes in zuul/.* The tooltip is hard to understand "1MB, 7/9" <- What is this ?

yes I tried to disable it but I think it is computing the sum of all the numbers below.

PR #138, opened 2020-04-25
preliminary work on a home page

flepied:
- web: redirect to the first index if there is one or display a message
- api: add a /indices route returning the list of names of the index
- db: use the monocle.changes. prefix for index name

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,7 @@ deps =
   -r test-requirements.txt
 
 [testenv:py3]
-commands = nosetests -v -s --nologcapture tests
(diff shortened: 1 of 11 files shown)


tox.ini
```diff
@@ -10,7 +10,7 @@ deps =
   -r test-requirements.txt
 
 [testenv:py3]
-commands = nosetests -v -s --nologcapture tests
```
morucci, line 13: This convenient why remove it ?

flepied: it was needed to debug the tests when they were failing.

PR #143, opened 2020-04-27
display Mean time to merge and commits per change in lifecycle stats

flepied:


morucci:
Looks good to me. I merge.

PR #146, opened 2020-04-29
web: display the timeline of commits in /change/

flepied:


web/src/components/commits_timeline.js
```diff
@@ -0,0 +1,50 @@
...
+  getTitle (commit) {
+    return commit.title
+  }
+
+  getDate (commit) {
+    return moment(commit.authored_at).fromNow() + ' by ' + commit.author
+  }
+
+  render () {
+    console.log(this.props.data)
```
morucci, line 32: Maybe just remove this console.log. I approve so let's merge after that removal.

PR #153, opened 2020-04-30
Display approval state to change listing

morucci:


monocle/github/pullrequest.py
```diff
@@ -336,6 +337,7 @@ def extract_pr_objects(pr):
             change['text'] = pr['bodyText']
             change['additions'] = pr['additions']
             change['deletions'] = pr['deletions']
+            change['approval'] = pr['reviewDecision']
```
flepied, line 340: we should make it a list like in gerrit  to make the code easier at the API level.

morucci: Done

morucci:
Ok I just rebased it

PR #172, opened 2020-05-05
Improve support for issue tracker links detection

morucci:
At indexation ensure text ant title field are also indexed as keyword
to enable regexp search.

In the web API after the EL query for a change enhances the change
with:

- has_issue_tracker_links field: choice 'generic|github.com|altassian.net'
- issue_tracker_links: The list of matchers + replacements

At query time add the has_issue_tracker_links params to scope
on changes having the tracker links.

Add unit tests.

For now only support generic tracker url, github.com and altassian.net

### DIFF
--- a/monocle/db/queries.py
+++ b/monocle/db/queries.py
@@ -99,17 +99,18 @@ def generate_changes_filter(params, qfilter):
             {"regexp": {"changed_files.path": {'value': Detector.tests_regexp}}}
         )
     if has_issue_tracker_links:
-        value = Detector().get_issue_tracker_regexp()
-        qfilter.append(
-            {
-                "bool": {
-                    "should": [
-                        {"regexp": {"text.keyword": {'value': value}}},
-                        {"regexp": {"title.keyword": {'value': value}}},
-                    ]
+        value = Detector().get_issue_tracker_regexp(style=has_issue_tracker_links)
+        if value:
+            qfilter.append(
+                {
+                    "bool": {
+                        "should": [
+                            {"regexp": {"text.keyword": {'value': value}}},
+                            {"regexp": {"title.keyword": {'value': value}}},
+                        ]
+                    }
                 }
-            }
-        )
+            )
 
 
 def generate_filter(repository_fullname, params):
--- a/monocle/db/queries.py
+++ b/monocle/db/queries.py
@@ -91,11 +91,24 @@ def generate_events_filter(params, qfilter):
 def generate_changes_filter(params, qfilter):
     state = params.get('state')
     tests_included = params.get('tests_included')
+    has_issue_tracker_links = params.get('has_issue_tracker_links')
     if state:
         qfilter.append({"term": {"state": state}})
     if tests_included:
         qfilter.append(
             {"regexp": {"changed_files.path": {'value': Detector.tests_regexp}}}
+        )
+    if has_issue_tracker_links:
+        value = Detector().get_issue_tracker_regexp()
+        qfilter.append(
+            {
+                "bool": {
+                    "should": [
+                        {"regexp": {"text.keyword": {'value': value}}},
+                        {"regexp": {"title.keyword": {'value': value}}},
+                    ]
+                }
+            }
         )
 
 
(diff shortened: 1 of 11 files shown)


monocle/db/queries.py
```diff
@@ -91,12 +91,25 @@ def generate_events_filter(params, qfilter):
 def generate_changes_filter(params, qfilter):
     state = params.get('state')
     tests_included = params.get('tests_included')
+    has_issue_tracker_links = params.get('has_issue_tracker_links')
     if state:
         qfilter.append({"term": {"state": state}})
     if tests_included:
         qfilter.append(
```
flepied, line 98: should we do the same thing here with a "should" request?

monocle/utils.py
```diff
@@ -54,18 +54,87 @@ class Detector(object):
     tests_regexp = ".*[Tt]est.*"
     tests_re = re.compile(tests_regexp)
 
+    issue_tracker_links = {
```
flepied, line 57: do we want to support other bug trackers like Jira or bugzilla? Enterprise GitHub?

monocle/db/queries.py
morucci: Why ? to match the test regexp in title or text ?

monocle/utils.py
morucci: Yes. We will be able to complete with others issue trackers in SAS mode like jira.altasian. In a second step the idea is to make the regexps configurable when we will have the project feature. The Detector will take a project config in input.  But let keep it simple for a first step.

flepied: I was thinking something simple like detect url with jira|bug|issue in them.

morucci: altassian.net would not match in that case.

monocle/db/queries.py
flepied: no but to have boolean result instead of the files.

monocle/utils.py
morucci: but I could probably add a generic url matcher too.

morucci: Hum at query level adding to much regexp (so increasing complexity) is forbidden by lucene:
```
"index":"monocle.test.monocle-unittest","caused_by":{"type":"too_complex_to_determinize_exception","reason":"Determinizing .*https?:\\/\\/.*issue.*|https?:\\/\\/.*bug.*|https?:\\/\\/.*jira.*|https?:\\/\\/.*toto.*|https?:\\/\\/.*titi.*.*|.* \\#[0-9]+.*|.*[^/ :]+\\/[^/]+\\#[0-9]+.*|.*GH-[1-9]+.*|.*https?:\\/\\/.+.atlassian.net\\/browse\\/.*.* would result in more than 10000 states.","caused_by":{"type":"too_complex_to_determinize_exception","reason":"Determinizing automaton with 111 states and 415 transitions would result in more than 10000 states."}}}}]},"status":400}
```

We should probably remove the has_issue_tracker_links query parameter. For the moment it does not make sense to support every style during a query. Nobody will have issue tracker on multiple tracker. So we could re-enable it when we will have the project definition support where we will be able to say my project: issue-tracker: github.com or altassian.net or whatever.

So my proposal is to keep that the change is enhanced with the has_issue_tacker_links attribute. But remove the has_issue_tracker_links query parameter.

What do you think ?

monocle/db/queries.py
morucci: Sorry I don't get it :(

monocle/utils.py
morucci: Or the has_issue_tracker_links query parameter could contains choices: generic, github.com, altassian.net So the number of regexp is limited to a known set.

morucci: I've implemented my last proposal to prevent having to complex regexp queries.

flepied: I think it's a good time to start having this in a specific section in the configuration with a regexp per index.

monocle/db/queries.py
flepied: ok let's discuss it irl.

flepied: understood

PR #173, opened 2020-05-05
[crawler] fix 2 errors discovered while indexing the Ansible org

flepied:
- fix GitHub PR with empty files
- sleep when 'You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.'


monocle/github/pullrequest.py
```diff
@@ -213,6 +214,9 @@ def _getPage(self, kwargs, prs):
             self.log.info("Total PRs to fetch: %s" % kwargs['total_prs_count'])
         if 'data' not in data:
             self.log.error('No data collected: %s' % data)
+            if 'message' in data and 'wait a few minutes' in data['message']:
+                self.log.debug('sleeping 120s')
```
morucci, line 218: Let's give a bit more information in the log message. I would have set it to info.

morucci: or warning

flepied: ok

PR #175, opened 2020-05-06
Abandoned pies

flepied:


morucci:
I get this when I'm trying to deploy:

```
Creating an optimized production build...
Failed to compile.

./src/App.js
Cannot find file './components/authors_abandoned_pie' in './src'.
```

monocle/db/queries.py
```diff
@@ -273,6 +271,7 @@ def _events_top(es, index, repository_fullname, field, params):
         "size": 0,
         "query": generate_filter(repository_fullname, params),
     }
+    print(body)
```
morucci, line 274: to remove

flepied: done

PR #178, opened 2020-05-06
Test ratio

flepied:


monocle/db/queries.py
```diff
@@ -433,6 +433,24 @@ def change_merged_avg_commits(es, index, repository_fullname, params):
     return data['aggregations']['agg1']['value']
 
 
+def changes_with_tests_ratio(es, index, repository_fullname, params):
+    params = deepcopy(params)
+    params['etype'] = ("Change",)
+    all = count_events(es, index, repository_fullname, params)
+    if all == 0:
+        return 0
+    params['files'] = Detector.tests_regexp
```
morucci, line 442: Why don't you use the tests_included params for this ?

monocle/db/queries.py
```diff
@@ -433,6 +433,24 @@ def change_merged_avg_commits(es, index, repository_fullname, params):
...
+    tests = count_events(es, index, repository_fullname, params)
+    return round(tests / all * 100, 1)
+
+
+def count_opened_changes(es, index, repository_fullname, params):
+    params = deepcopy(params)
+    params['etype'] = ("Change",)
+    params['state'] = "OPEN"
+    return count_events(es, index, repository_fullname, params)
```
morucci, line 451: We should rename count_events to count_objects. But I fine with this.

monocle/db/queries.py
flepied: done

PR #188, opened 2020-05-08
web: Unify error handling for the changes components

morucci:


morucci:
@flepied Why do we need that special handling just at those places ? This PR fixes an unexpected error message where the data is empty (normal behaviour) as the filter is set on an unknown user (in my test).
I think props.data must be as expected because query is no longer loading neither in an error. If props.data.items is an empty list that's also fine. So maybe something I don't get there. Let me know.

flepied:
> @flepied Why do we need that special handling just at those places ? This PR fixes an unexpected error message where the data is empty (normal behaviour) as the filter is set on an unknown user (in my test).
> I think props.data must be as expected because query is no longer loading neither in an error. If props.data.items is an empty list that's also fine. So maybe something I don't get there. Let me know.

ok let's got like that

PR #191, opened 2020-05-09
use uwsgi to serve API

flepied:


monocle/webapp.py
```diff
@@ -150,14 +161,6 @@ def indices():
 
 
 def main():
-    config_path = os.getenv('CONFIG', None)
```
morucci, line 153: Why not keeping this as a function. I'd like to have the authorization support when I run the flask dev server outside of docker or w/o the WSGI server

doc/adr/0006-cache-es-queries.md
```diff
@@ -0,0 +1,38 @@
...
+distributed solution if needed in the future.
+
+## Considered Options
+
+* [option 1] do caching in front of the API service with a reverse proxy.
+* [option 2] do caching inside the API service.
+
+## Decision Outcome
+
+Chosen option: "[option 1]", because we have an optional
```
morucci, line 26: we choose option 2 instead.

monocle/webapp.py
flepied: I needed to have it called when loaded in all the cases. I don't understand what will be the advantage of having it in a function.

doc/adr/0006-cache-es-queries.md
flepied: good catch!

monocle/webapp.py
morucci: ok that's fine yes.

PR #215, opened 2020-05-23
github: enable use of a GitHub app for crawler tokens

morucci:


morucci:
@lepture thanks for the heads-up: https://github.com/change-metrics/monocle/pull/235

PR #218, opened 2020-05-27
adaptative retries for github queries

flepied:


monocle/github/graphql.py
```diff
@@ -86,7 +90,10 @@ def getRateLimit(self):
         return data['data']['rateLimit']
 
     @retry(
-        before=before_log(log, logging.INFO), wait=wait_exponential(min=10, max=300),
+        after=after_log(log, logging.INFO),
+        wait=wait_exponential(),
+        stop=stop_after_attempt(3),
+        reraise=True,
     )
     def query(self, qdata, skip_get_rate_limit=False, ignore_not_found=False):
```
morucci, line 98: ignore_not_found is not used

monocle/github/pullrequest.py
```diff
@@ -21,6 +21,12 @@
 from dataclasses import dataclass
 from pprint import pprint
 
+from .graphql import Timeout
+
+MAX_TRY = 3
+MAX_BULK_SIZE = 100
```
morucci, line 27: I was not lucky at all with so high bulk size.

monocle/github/graphql.py
```diff
@@ -86,7 +90,10 @@ def getRateLimit(self):
         return data['data']['rateLimit']
 
     @retry(
-        before=before_log(log, logging.INFO), wait=wait_exponential(min=10, max=300),
+        after=after_log(log, logging.INFO),
+        wait=wait_exponential(),
+        stop=stop_after_attempt(3),
```
morucci, line 95: If I understand well, this 3 failures before we will manage to reduce the bulk size in the caller ?

flepied: yes because it seems to depend on the machine answering: sometime it's working on another attempt

monocle/github/pullrequest.py
flepied: that's the max and it works from time to time but we adapt the size dynamically.

monocle/github/graphql.py
flepied: ok removing it

monocle/github/pullrequest.py
```diff
@@ -226,32 +235,56 @@ def _getPage(self, kwargs, prs):
...
             )
             try:
                 hnp = self._getPage(kwargs, prs)
+                kwargs['size'] = min(MAX_BULK_SIZE, int(kwargs['size'] * AUGMENT) + 1)
+                one = 0
+                if not get_commits:
+                    self.log.info('Will get full commits on next query.')
+                    kwargs['pr_query'] = self.pr_query % self.pr_commits
+                    get_commits = True
             except requests.exceptions.ConnectionError:
```
morucci, line 262: We don't need this with tenacity I think.

monocle/github/graphql.py
morucci: ok

monocle/github/pullrequest.py
morucci: ok, but maybe not so high because we will have a high time-out rate that will slow down the crawling process a lot.

monocle/github/graphql.py
morucci: stop_after_attempt=2 is enough

monocle/github/graphql.py
```diff
@@ -86,9 +90,12 @@ def getRateLimit(self):
         return data['data']['rateLimit']
 
     @retry(
-        before=before_log(log, logging.INFO), wait=wait_exponential(min=10, max=300),
+        after=after_log(log, logging.INFO),
+        wait=wait_fixed(1),
```
morucci, line 94: What about using retry_unless_exception_type('Timeout')  because if we have a HTTP timeout or other transport/HTTP errors we might need to retry and even with wait_exponential. We don't want to retry the same query when we have that API timeout we let the caller rework the size.

PR #229, opened 2020-06-09
[web] Refactored mergeability status into a badge

morucci:


web/src/components/common.js
```diff
@@ -110,14 +110,39 @@ function chooseApprovalBadgeStyle (app, idx = 0) {
...
     if (this.props.data.state === 'OPEN' && this.props.data.draft) {
       return <Badge variant="dark">Draft</Badge>
     }
     switch (this.props.data.state) {
       case 'OPEN':
-        return <span><Badge variant="success">Open</Badge> <small>[{this.props.data.mergeable.toLowerCase()}]</small></span>
+        return (
+          <span>
+            <Badge variant="success">Open</Badge>{' '}
+            <MergeableStatusBadge mergeable={this.props.data.mergeable} />
```
flepied, line 143: you removed the lowercase call. I don't remember why it was there...

morucci: Yes the removal is intentional.

flepied: ok

PR #235, opened 2020-06-12
[crawler] github application: use authlib.jose.jwt instead of pyjwt

morucci:


morucci:
one less dependency because it comes with authlib that we already have in our requirements.txt file.

PR #239, opened 2020-06-19
Add gh workflow and fix queries related tz issues

morucci:


.github/workflows/units.yaml
```diff
@@ -0,0 +1,38 @@
+name: Monocle CI
+on: push
+
+jobs:
+  runner-job:
+    runs-on: ubuntu-latest
```
flepied, line 6: why ubuntu?

morucci: Because of no choice: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

PR #241, opened 2020-06-21
[crawler] compute if a change is self merged

flepied:


web/src/components/changes_lifecycle.js
```diff
@@ -185,23 +185,26 @@ class ChangesLifeCycleStats extends BaseQueryComponent {
...
-                      </ListGroup.Item>
                       <ListGroup.Item>
                         <Link to={`/${this.props.index}/changes${this.getSearchString('CLOSED')}`}>{data.abandoned} changes abandoned: {data.ratios['abandoned/created']}%</Link>
                       </ListGroup.Item>
                       <ListGroup.Item>
                         <Link to={`/${this.props.index}/changes${this.getSearchString('MERGED')}`}>{data.merged} changes merged: {data.ratios['merged/created']}%</Link>
                       </ListGroup.Item>
                       <ListGroup.Item>
-                        Changes with tests: {data.tests}%
+                        <Link to={`/${this.props.index}/changes${this.getSearchString('SELF-MERGED')}`}>{data.self_merged} changes self merged</Link>
```
flepied, line 195: we should put the percentage to be consistent with the other stats.

PR #284, opened 2021-01-12
Update git url from morruci to change-metrics

mzjp2:
I suspect this is a remnant from a rename or migrating the repository to a organisation. It's still technically correct (in that it works), but perhaps best to change it for full clarity.

Feel free to disregard and close the PR if you disagree or think it's too small a change. :)


morucci:
Hi Zain,

Good catch! thanks!

Issue #291, opened 2021-01-31
gerrit repository with / makes data dumping fail


flepied:
```ShellSession
$ python ./monocle/gerrit/review.py --base-url https://review.rdoproject.org/r/ --repository openstack/scciclient-distgit --id 31697 --output-dir /tmp
Traceback (most recent call last):
  File "./monocle/gerrit/review.py", line 412, in <module>
    json.dump(review[0], open(basepath + "_raw.json", "w"), indent=2)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/https:__review.rdoproject.org_r_-openstack/scciclient-distgit-31697_raw.json'
```

PR #292, opened 2021-01-31
gerrit: allow to prefix projects and small fixes

flepied:
- allow to prefix projects by passing a prefix setting in the config file
- fix dumping of projects with / (bug #291)
- fix dumping data with cli (bug #293)

### DIFF
--- a/monocle/gerrit/review.py
+++ b/monocle/gerrit/review.py
@@ -1,5 +1,5 @@
 # Monocle.
-# Copyright (C) 2019-2020 Monocle authors
+# Copyright (C) 2019-2021 Monocle authors
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published
@@ -364,8 +364,9 @@ class ReviewesFetcher(object):
 
 
 if __name__ == "__main__":
-    import os
     import argparse
+    import dataclasses
+    import os
     from pprint import pprint
 
     parser = argparse.ArgumentParser(prog="review")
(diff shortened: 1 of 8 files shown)


monocle/gerrit/review.py
```diff
@@ -116,6 +124,9 @@ def get(self, updated_since, change=None):
                     break
             else:
                 break
+        if self.prefix:
+            for review in reviews:
+                review["project"] = self.prefix + review["project"]
```
morucci, line 129: Line 165 and 167 the computation for repository_prefix and repository_shortname won't work as expected for prefix such as "rdo/" as we split on "/".

morucci, line 129: So the repository_shortname and repository_prefix line 165 and 167 need to be fixed as we split on "/" so for prefix suc as "rdo/" it won't be as expected.

tox.ini
```diff
@@ -11,6 +11,8 @@ deps =
 
 [testenv:py3]
 commands = nosetests -v monocle/tests
+         python ./monocle/gerrit/review.py --base-url https://gerrit-review.googlesource.com --repository gerrit --id 246332 --output-dir /tmp
```
morucci, line 14: Why is this needed ? Why not provide the raw.json and extracted.json in fixtures ?

morucci: I remember we used to lose these commands. I think we should add a new tox command to add all the subcommand to fetch github and gerrit fixtures. This way we keep the commands, and it is on demand. Here I don't this that's ok to fetch the data when running the test suite.

morucci (changes requested):

PR #296, opened 2021-01-31
Add BaseCrawler base class to dereivated by crawler

morucci:


flepied (commented):
need a CHANGELOG entry but lgtm

PR #298, opened 2021-02-04
Add instructions to run the python tests

TristanCacqueray:
This change documents how to run the python tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,23 @@ To use `docker-compose` from source code, use the
 $ ln -sf docker-compose.yml.dev docker-compose.yml
 ```
 
+## To run the tests
+
+Tests rely on the Elasticsearch service so first you need to ensure the Elasticsearch container is running.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -14,6 +14,21 @@ To use `docker-compose` from source code, use the
 $ ln -sf docker-compose.yml.dev docker-compose.yml
 ```
 
+## To run the tests
+
+To run the Python test, make sure an elasticsearch service is running on localhost,
+for example by executing:
+
```
morucci, line 21: I think we should tell, that:

Tests rely on the Elasticsearch service so first you need to ensure the Elasticsearch container is running. You can run the "docker-compose up -d" or you can only run the Elasticsearch container by running "docker-compose start elastic".

morucci (changes requested):

CONTRIBUTING.md
TristanCacqueray: Ok, do you mind keeping the podman command for user without docker?

morucci: +1

morucci (approved):

PR #300, opened 2021-02-11
Median Time To Merge

morucci:


requirements.txt
```diff
@@ -10,3 +10,4 @@ Flask-Caching
 uwsgi
 iso8601
 dacite
+cryptography<=3.3.2
```
flepied, line 13: why do we need this? could you add a comment?

flepied (approved):

PR #311, opened 2021-03-02
Fix example config

danpawlik:
The current example is **raising** an error in monocle crawler
service.


morucci (approved):
Good catch thx !

PR #309, opened 2021-02-25
Add support for podman-compose; remove PWD in volume path

danpawlik:
With this commit, Monocle project will be able to start with
podman-compose tool.

PR #318, opened 2021-03-10
Change Elasticsearch backend to Opendistro for Elasticsearch

danpawlik:
The new version of Elasticsearch has changed the License to
Elastic License 2.0 which is not very welcome in Opensource
world.
In that case, the best option is to switch the Elasticsearch to
Opendistro for Elasticsearch which is running under Apache 2.0 license.


danpawlik:
Base on: https://github.com/change-metrics/monocle/pull/309

monocle/db/db.py
```diff
@@ -174,13 +174,38 @@ def __init__(
         create=True,
         previous_schema=False,
         idents_config: Optional[IdentsConfig] = None,
+        user=None,
```
morucci, line 177: So a support for the authentication on the Elastic backend is fine to me. Would you mind to open a single PR with this support in the Monocle API and crawler (only @ python level). No need to provide specific unit test for that.

helpers/elastic.sh
```diff
@@ -0,0 +1,12 @@
...
+
+ELASTIC_OPENDISTRO=$(echo $ELASTIC_OPENDISTRO | tr '[:upper:]' '[:lower:]')
+
+if [ "$ELASTIC_OPENDISTRO" = "true" ]; then
+    # FIXME: Remove -k for checking TLS verification. Also remove admin:admin credentials.
+    curl --silent --fail -k \
+      --user ${ELASTIC_USER:-'admin'}:${ELASTIC_PASSWORD:-'admin'} \
+      https://localhost:9200/_cluster/health
+else
+    curl --silent --fail localhost:9200/_cluster/health
```
morucci, line 11: No issue removing the "| exit 1" ?

helpers/elastic.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+ELASTIC_OPENDISTRO=$(echo $ELASTIC_OPENDISTRO | tr '[:upper:]' '[:lower:]')
+
```
morucci, line 4: It could probably be simplified by something like:

```
auth=""
if [ "$ELASTIC_OPENDISTRO" = "true" ]; then
  auth=" --user ${ELASTIC_USER:-'admin'}:${ELASTIC_PASSWORD:-'admin'}"
fi
curl --silent $auth --fail localhost:9200/_cluster/health
```

helpers/monocle.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+ELASTIC_OPENDISTRO=$(echo $ELASTIC_OPENDISTRO | tr '[:upper:]' '[:lower:]')
+
+if [ "$ELASTIC_OPENDISTRO" = "true" ]; then
+    # FIXME: Remove --insecure option for checking TLS in the future.
+    #        Also change admin:admin credentials.
+    monocle --elastic-conn elastic:9200 \
```
morucci, line 8: ditto for the simplification.

docker-compose.yml.img
```diff
@@ -35,26 +37,30 @@ services:
...
     volumes:
-      - $PWD/etc:/etc/monocle:z
+      - ./etc:/etc/monocle:z
     healthcheck:
         test: python -c "import requests,sys; r=requests.get('http://localhost:9876/api/0/health'); print(r.text); sys.exit(1) if r.status_code!=200 else sys.exit(0)"
 
+  # ELASTIC_CONTAINER_URL is defined in .env file.
   elastic:
-    image: "docker.elastic.co/elasticsearch/elasticsearch:7.10.1"
+    image: ${ELASTIC_CONTAINER_URL}
```
morucci, line 46: I think we need to keep the default to elasticsearch image if no .env file provided.

docker-compose.yml.img
```diff
@@ -25,6 +25,8 @@ services:
     ports:
       - "${MONOCLE_API_ADDR:-0.0.0.0}:9876:9876"
       - "${MONOCLE_API_ADDR:-0.0.0.0}:9877:9877"
+    env_file:
+      - .env
```
morucci, line 29: Why it become mandatory. Could we keep tat file optional ?

morucci (changes requested):
So first thanks for that PR.

I'm not against adding support for OpenDistro but in the mean time I'd like to avoid complexity to manage the backend.

I understand that some would prefer to avoid the official elastic release when using Monocle so we need to ease that. To do so we need to ensure Monocle integrate well whatever the Elastic distribution.

I would like to propose some change to your proposal:

1. Could you extract the Monocle API/Crawler change to support Elastic connection authentication as well command line argument parse in main.py into a separate PR. We will merge it first. (add a changelog entry please)

2. Could ensure this PR is rebased on master and remove mixup with the other PR about podman ?

3. I'd like to keep the need for the .env optional. For a local test deployment it should not be needed.

TristanCacqueray:
Is this still relevant, and/or should this includes OpenSearch?

morucci:
I don't see this as a priority. As long as we do not expose EL API directly like EL as a Service we should be right regarding the License. What do you think @TristanCacqueray is moving to OpenDistro a right move ?
Even if we decide to move to OpenDistro this could be done later after we release version 1.

TristanCacqueray:
I think that is an implementation detail.

TristanCacqueray:
Let's re-open when needed.

PR #325, opened 2021-03-25
Added section for starting Monocle service without compose tool

danpawlik:
This commit provide manual, how to start Monocle service for developing
purpose.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,3 +93,81 @@ and making it executable with `chmod +x .git/hooks/pre-commit`.
 
 Please refer to the file `dummy/change.py` that contain a Dummy driver and
 some comments to help you get started.
+
+## Use docker/podman commands for start Monocle
+
+Below commands will start Monocle service without using compose tool.
+Mounting the project in a Python library path, will be also helpful.
+
+```Shell
+# Set public IP address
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -93,3 +93,81 @@ and making it executable with `chmod +x .git/hooks/pre-commit`.
 
 Please refer to the file `dummy/change.py` that contain a Dummy driver and
 some comments to help you get started.
+
+## Use docker commands for start Monocle
```
morucci, line 97: Use docker/podman ...

CONTRIBUTING.md
```diff
@@ -93,3 +93,81 @@ and making it executable with `chmod +x .git/hooks/pre-commit`.
...
+
+The easiest way for developing Monocle project is to run Monocle without
```
morucci, line 99: Not agree with that. I'm using compose to develop following that CONTRIBUTING guide and that's fine :)
Please remove those sentences.

Instead could you add a warning by telling that this process is for informational only. It is here for folks that don't have  or don't want to use docker-compose.

morucci (changes requested):

morucci:
@danpawlik Please let me know what do you think of https://github.com/change-metrics/monocle/pull/327

morucci:
I'm merging it then I'll merge 327 as it works as expected.

PR #326, opened 2021-03-29
Added project_definition support

danpawlik:
This commit provides project_definition support. It means
that you are able to define project in your index definition,
and monocle will return only informations related to
repositories-regex [1]. For example:

tenants:
  - index: openstack
    crawler:
      loop_delay: 600
      gerrit_repositories:
        - name: "^openstack/.*"
          updated_since: "2021-01-01"
          base_url: https://review.opendev.org/
    projects:
      - name: infra
        repositories-regex: "opendev/system-config|openstack/disskimage-builder"

The infra project definition returns statistics related only for those
repositories that are in repositories-regex.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-regexp-query.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ Monocle is able to index changes from multiple code review systems. A contributo
 might get different identities across code review systems. Thus Monocle provides
 a configuration section to define aliases for contributors.
 
-Let say a Monocle index is configured to fetch changes from github.com and review.opendev.org (Gerrit) and we would like that John's metrics are merged under the `John Doe` identity. 
+Let say a Monocle index is configured to fetch changes from github.com and review.opendev.org (Gerrit) and we would like that John's metrics are merged under the `John Doe` identity.
 
 ```YAML
 tenants:
--- a/monocle/config.py
+++ b/monocle/config.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+from dataclasses import dataclass
 from typing import Dict, List
 from jsonschema import validate as schema_validate
 from jsonschema import draft7_format_checker
@@ -136,6 +137,17 @@ schema = {
                             },
                         },
                     },
+                    "projects": {
+                        "description": "Project definition",
+                        "type": "array",
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "name": {"type": "string"},
+                                "repositories_regex": {"type": "string"},
+                            },
+                        },
+                    },
                 },
             },
         }
@@ -189,6 +201,9 @@ tenants:
           login: fabien
           password: secure
           prefix: namespace/
+    projects:
+      - name: infra
+        repositories_regex: "^config|infra$|openstack/.*"
 """
 
 
@@ -202,6 +217,12 @@ def validate(data, schema):
 
 class Username(str):
     pass
+
+
+@dataclass
+class ProjectDefinition(str):
+    name: str
+    repositories_regex: str
 
 
 def build_index_acl(config: dict) -> Dict[str, List[Username]]:
@@ -235,3 +256,13 @@ def get_idents_config(config: dict, index_name: str) -> IdentsConfig:
         return list(map(ident_from_config, matches[0].get("idents", [])))
     else:
         return []
+
+
+def build_project_definitions(config: dict) -> Dict[str, List[ProjectDefinition]]:
+    indexes_project_def: Dict[str, List[ProjectDefinition]] = {}
+    for tenant in config["tenants"]:
+        if "projects" not in tenant.keys():
+            indexes_project_def[tenant["index"]] = []
+        else:
+            indexes_project_def[tenant["index"]] = tenant["projects"]
+    return indexes_project_def
--- a/monocle/tests/unit/test_api.py
+++ b/monocle/tests/unit/test_api.py
@@ -16,7 +16,9 @@
 
 import logging
 import os
+import tempfile
 import unittest
+import yaml
 from flask import json
 from monocle import webapp
 from monocle import config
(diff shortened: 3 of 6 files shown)


monocle/webapp.py
```diff
@@ -121,11 +121,37 @@ def whoami():
...
-@app.route("/api/0/query/<name>", methods=["GET"])
-def query(name):
+def _get_index(request):
     if not request.args.get("index"):
         abort(make_response(jsonify(errors=["No index provided"]), 404))
-    index = request.args.get("index")
+    return request.args.get("index")
+
+
+def _get_project_definition(index):
```
morucci, line 130: You should read the ProjectConfig the same way than in line 81. And put the data in global (yes that's bad, there is a single place on that file where we read the configuration file so better to not duplicate).

monocle/webapp.py
```diff
@@ -121,11 +121,37 @@ def whoami():
...
+    return (
+        config.get_project_definition(monocle_config, index) if monocle_config else None
+    )
+
+
+@app.route("/api/0/config_project_def", methods=["GET"])
+def get_cfg_project_definition():
+    index = _get_index(request)
+    projects = _get_project_definition(index)
+    if not projects:
```
morucci, line 141: If no project definition then simply return an empty list

monocle/webapp.py
```diff
@@ -121,11 +121,37 @@ def whoami():
...
+
+
+def _get_project_definition(index):
+    monocle_config = yaml.safe_load(open(config_path)) if config_path else None
+    return (
+        config.get_project_definition(monocle_config, index) if monocle_config else None
+    )
+
+
+@app.route("/api/0/config_project_def", methods=["GET"])
```
morucci, line 137: I prefer such simply endpoint name.
@app.route("/api/0/projects", ...

morucci: Also this need to be protected by the ACL

monocle/webapp.py
```diff
@@ -152,8 +178,17 @@ def query(name):
...
     params = utils.set_params(args)
+
+    if params["project_definition"] and not projects:
+        return (
+            "There are no project definion set in "
+            "config file: %s" % request.get("index"),
+            404,
+        )
+
+    params["config_project_definitions"] = projects
```
morucci, line 191: Not sure why we need that. We already have the endpoint that is api/0/projects so that's enough.

monocle/webapp.py
```diff
@@ -121,11 +121,37 @@ def whoami():
...
+            404,
+        )
+    return jsonify(projects)
+
+
+@app.route("/api/0/query/<name>", methods=["GET"])
+def query(name):
+
+    index = _get_index(request)
+    projects = _get_project_definition(index)
```
morucci, line 153: Not needed.

monocle/utils.py
```diff
@@ -208,6 +208,7 @@ def getter(attr, default):
     params["has_issue_tracker_links"] = getter("has_issue_tracker_links", None)
     params["change_ids"] = getter("change_ids", None)
     params["target_branch"] = getter("target_branch", None)
+    params["project_definition"] = getter("project_definition", None)
```
morucci, line 211: Not needed.

monocle/db/queries.py
```diff
@@ -72,6 +72,7 @@
     "new_contributors",
```
morucci, line 72: I don't see the need to change something in that file.

monocle/tests/unit/test_api.py
```diff
@@ -140,3 +142,86 @@ def test_whoami(self):
             sess["username"] = "jane"
         resp = self.client.get("/api/0/whoami")
         self.assertEqual("jane", json.loads(resp.data))
+
```
morucci, line 145: Only the api/0/projects endpoint must be validated.

morucci (changes requested):
I think we only need the config schema + the api/0/projects endpoint + the unittests for the endpoint. I don't get the changes in the db/queries.py.

The UI will get a component that query this new endpoint to get projects definitions for a given index.

monocle/webapp.py
danpawlik: I really don't like. Putting something as a global is going to raise some problems, but as you prefer.

monocle/webapp.py
danpawlik: Then in react will complicate next pull request, where axios will get array with promise object that it will comes to undefined and it will other issues. Can not be like it is now?

monocle/config.py
```diff
@@ -204,6 +219,10 @@ class Username(str):
     pass
 
 
+class ProjectDefinition(str):
```
morucci, line 222: @dataclass
class ProjectDefinition:
    name: str
    repositories_regex: str

monocle/db/queries.py
```diff
@@ -1023,3 +1024,21 @@ def repos_summary(es, index, repository_fullname, params):
         )
         repos[name]["changes_merged"] = count_merged_changes(es, index, name, params)
     return {"summary": repos}
+
+
```
morucci, line 1028: Not needed

monocle/webapp.py
morucci: ok

monocle/webapp.py
```diff
@@ -154,6 +177,16 @@ def query(name):
...
+    projects = project_defs[index] if index in project_defs else []
+
+    if params["project_definition"] and not projects:
+        return (
+            "There are no project definion set in "
+            "config file for : %s" % params["project_definition"],
+            404,
+        )
+
+    params["config_project_definitions"] = projects
```
morucci, line 189: Not needed

morucci (changes requested):

monocle/webapp.py
```diff
@@ -154,6 +177,19 @@ def query(name):
...
+        return (
+            "There are no project definion set in "
+            "config file for : %s" % params["project_definition"],
+            404,
+        )
+
+    for project_def in projects:
+        if project_def.get("name") == params.get("project_definition"):
+            repository_fullname = project_def.get("repositories_regex")
+
```
morucci, line 192: Not needed.

morucci (changes requested):
We don't need a new query parameter "project_definition". All parameters to filter based on a project definition already exists.

monocle/tests/unit/test_api.py
```diff
@@ -140,3 +142,88 @@ def test_whoami(self):
...
+            )
+            resp_data = json.loads(resp.data)
+            self.assertEqual(1, resp_data["total"])
+
+            # set wrong index
+            fake_index = "some-fake-index"
+            resp = self.client.get(
+                "/api/0/query/changes?index=%s&project_definition=test" % fake_index
+            )
+            self.assertEqual(404, resp.status_code)
```
morucci, line 189: This test is not needed.

morucci (changes requested):
See inline comments.

README.md
```diff
@@ -378,6 +378,16 @@ To check that, you are able to run simple curl command:
...
+## Using API
+
+Below there are describe few examples, how to use API in Monocle:
+
+```
+curl -XGET "http://localhost:9876/api/0/query/last_changes?index=zuul&repository=.*&gte=2020-12-24&from=0&size=100&state=OPEN&selfMerged=false&ec_same_date=true"
+curl -XGET "http://localhost:9876/api/0/projects?index=openstack"
+curl -XGET "http://localhost:9876/api/0/query/changes?index=monocle&project_definition=test"
+```
+
```
morucci, line 390: I think this should be handled in another PR dedicated to improve the documentation for the API usage.

morucci (changes requested):
That's look better IHMO now :) Just few things:
- Could you move the documentation of the API in another PR. This will be a place holder to have a more complete API definition doc in the README.md ?
- Could you add a Changelog entry ?

Then I'll merge the change.

morucci (approved):

PR #330, opened 2021-04-07
Added project definition filtering on Monocle web page

danpawlik:
With this commit, users will be able to filter the
results by project definition.

### DIFF
--- a/web/src/api.js
+++ b/web/src/api.js
@@ -83,9 +83,22 @@ function getLoggedUser () {
     })
 }
 
+function getConfigProjectDefinitions(pathname) {
+  /* Return json list that include project definitions from config file */
+  const indexname = pathname.replace("/", "");
+  const url = baseurl + "/projects?index=" + indexname;
+  return axios
+    .get(url)
+    .then((response) => response.data)
+    .catch((response) =>
+      console.log(`None object found on doing query: ${url}`)
+    );
+}
+
 export {
   getQueryResults,
   getIndices,
   getLoggedUser,
-  baseurl
+  baseurl,
+  getConfigProjectDefinitions
 }
--- a/web/src/components/filtersform.js
+++ b/web/src/components/filtersform.js
@@ -30,6 +30,7 @@ import { withRouter } from 'react-router-dom'
 import DatePicker from 'react-datepicker'
 import 'react-datepicker/dist/react-datepicker.css'
 import moment from 'moment'
+import { getConfigProjectDefinitions } from '../api'
 
 class DateFormBox extends React.Component {
   constructor (props) {
@@ -181,7 +182,9 @@ class FiltersForm extends React.Component {
       approvals: '',
       excludeApprovals: '',
       state: 'open',
-      selfApproved: false
+      selfApproved: false,
+      selectedProjectDef: '',
+      project_defs: []
     }
   }
 
@@ -190,6 +193,7 @@ class FiltersForm extends React.Component {
     this.unlisten = this.props.history.listen((location, action) => {
       this.fetchQueryParams()
     })
+    this.fetchProjectDefs()
   }
 
   componentWillUnmount () {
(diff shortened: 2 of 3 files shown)


danpawlik:
Related to https://github.com/change-metrics/monocle/pull/326

web/src/components/filtersform.js
```diff
@@ -276,15 +280,37 @@ class FiltersForm extends React.Component {
...
     </b>
     return resume
   }
+  handleProjectDefState = (e) => {
+    const selected = e.target.attributes.value.value;
+    this.setState({ project_definition: selected });
+  };
+
+  renderProjectDefs = async () => {
+    let res = await getConfigProjectDefinitions(this.props.location.pathname);
```
TristanCacqueray, line 294: Using await here seems unnecessary. Perhaps adding another then would work better, e.g.:  `getConfigProjectDefinitions(..).then(res => this.setState(..)`

web/src/components/filtersform.js
```diff
@@ -276,15 +280,38 @@ class FiltersForm extends React.Component {
...
       {'.'}
     </b>
     return resume
   }
+  handleProjectDefState = (e) => {
+    const selected = e.target.attributes.value.value;
+    this.setState({ project_definition: selected });
+  };
+
+  renderProjectDefs = () => {
```
TristanCacqueray, line 293: While this may work, I don't think storing the dom elements in the state is a good design. Perhaps it would be better to simply save the response with setState here, and do the rendering in the render function below.

web/src/components/filtersform.js
```diff
@@ -276,15 +280,38 @@ class FiltersForm extends React.Component {
...
+    getConfigProjectDefinitions(this.props.location.pathname).then((res) =>
+      this.setState({
+        ProjectDefs: res.map((response) => (
+          <Dropdown.Item
+            key={response.name}
+            value={response.name}
+            active={response.name === this.state.state}
```
TristanCacqueray, line 300: I think you need an extra state property to store the selected project, `this.state.state` seems odd.

danpawlik:
@TristanCacqueray hey, could you review once again?

web/src/components/filtersform.js
```diff
@@ -179,7 +180,8 @@ class FiltersForm extends React.Component {
       approvals: '',
       excludeApprovals: '',
       state: 'open',
-      selfApproved: false
+      selfApproved: false,
```
TristanCacqueray, line 183: perhaps add a default `project_defs: []` ?

web/src/components/filtersform.js
```diff
@@ -188,6 +190,7 @@ class FiltersForm extends React.Component {
     this.unlisten = this.props.history.listen((location, action) => {
       this.fetchQueryParams()
     })
+    this.renderProjectDefs();
```
TristanCacqueray, line 193: nit: `fetchProjectDefs` sounds like a better name

web/src/components/filtersform.js
```diff
@@ -179,7 +180,8 @@ class FiltersForm extends React.Component {
       approvals: '',
       excludeApprovals: '',
       state: 'open',
-      selfApproved: false
+      selfApproved: false,
+      project_definition: ""
```
TristanCacqueray, line 184: is this `selected_project` ?

web/src/components/filtersform.js
```diff
@@ -280,11 +284,36 @@ class FiltersForm extends React.Component {
     </b>
     return resume
   }
+  handleProjectDefState = (e) => {
+    const selected = e.target.attributes.value.value;
+    this.setState({ repository: selected })
```
TristanCacqueray, line 289: It seems like this should be: `selectedProjectDef` instead of `repository`

web/src/api.js
```diff
@@ -83,9 +83,22 @@ function getLoggedUser () {
     })
 }
 
+function getConfigProjectDefinitions(pathname) {
+  /* Return json list that include project definitions from config file */
+  const indexname = pathname.replace("/", "");
+  const url = baseurl + "/projects?index=" + indexname;
+  return axios
+    .get(url)
+    .then((response) => response.data)
```
TristanCacqueray, line 92: The `api.js` module does not process the response or catch error in the other function. It seems better to push that work downstream, is there a reason `getConfigProjectDefinitions` handle the axios promise?

web/src/api.js
```diff
@@ -83,9 +83,22 @@ function getLoggedUser () {
...
+    .catch((response) =>
+      console.log(`None object found on doing query: ${url}`)
```
TristanCacqueray, line 94: nit: when hiding an exception, use console.error, e.g.:
```javascript
} catch (ex) {
  console.error(`${url} failed`, ex.message);
}
```

web/src/components/filtersform.js
```diff
@@ -397,6 +425,23 @@ class FiltersForm extends React.Component {
                                 placeholder="Files regexp"
                                 onChange={v => this.handleChange('files', v)}
                               />
+                              </Form.Group>
+                            <Form.Group controlId="formProjectDefinitionInput">
```
TristanCacqueray, line 429: Perhaps it would be better to wait for https://github.com/change-metrics/monocle/pull/341 and rebase on the new FilterForm component.

TristanCacqueray (changes requested):

web/src/components/filtersform.js
TristanCacqueray: Note that this PR is now rebased on the new patternfly component.

morucci:
@danpawlik thanks again ! I'm closing this one, as it have been implemented on top of rescript code.

PR #335, opened 2021-04-09
Validate Javascript files with Prittier.io tool

danpawlik:
The Prittier.io tool is validating if the Javascript files
are nice to read.


danpawlik:
Do we want to switch to eslint + prettier or just eslint is enough?

web/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+    "semi": true,
```
morucci, line 2: https://prettier.io/docs/en/options.html#semicolons
I'd like to have it at false. It seems semicolons can be omitted and be added by the JS engine before compile.

morucci (changes requested):
Hi, Thanks I'm ok with your change, except I don't want the semicolon. Could you rework it with the option at False.

morucci:
Thanks for the change, I'll probably merge this rescript PR first:  https://github.com/change-metrics/monocle/pull/335 then we can rebase this one on it :)

TristanCacqueray:
Daniel has been working on this for longer than the rescript change, we could merge it first.

morucci:
@TristanCacqueray Alright let's merge it now :) Thanks Daniel !!!

morucci (approved):

PR #336, opened 2021-04-10
Add Kubernetes Resources

guy9050:
Make it possible to deploy monocle on k8s/openshift

### DIFF
--- a/k8s/monocle-deployment.yml
+++ b/k8s/monocle-deployment.yml
@@ -1,0 +1,287 @@
+apiVersion: v1
+kind: List
+items:
+  - apiVersion: v1
+    kind: PersistentVolumeClaim
+    metadata:
+      creationTimestamp: null
+      labels:
+        io.kompose.service: config-data
+      name: config-data
+    spec:
+      accessModes:
+        - ReadWriteMany
+      resources:
+        requests:
+          storage: 100Mi
+    status: { }
+  - apiVersion: v1
+    kind: PersistentVolumeClaim
+    metadata:
+      creationTimestamp: null
+      labels:
+        io.kompose.service: elastic-data
+      name: elastic-data
+    spec:
+      accessModes:
+        - ReadWriteOnce
+      resources:
+        requests:
+          storage: 100Mi
+    status: { }
+  - apiVersion: v1
+    kind: PersistentVolumeClaim
+    metadata:
+      creationTimestamp: null
+      labels:
+        io.kompose.service: crawler-dump-data
+      name: crawler-dump-data
+    spec:
+      accessModes:
+        - ReadWriteOnce
+      resources:
+        requests:
+          storage: 100Mi
+    status: { }
+  - apiVersion: v1
+    kind: Service
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert
+        kompose.version: 1.17.0 (HEAD)
+      creationTimestamp: null
+      labels:
+        io.kompose.service: api
+      name: api
+    spec:
+      ports:
+        - name: http-rest
+          port: 9876
+          targetPort: 9876
+        - name: websocket
+          port: 9877
+          targetPort: 9877
+      selector:
+        io.kompose.service: api
+    status:
+      loadBalancer: { }
+  - apiVersion: v1
+    kind: Service
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert
+        kompose.version: 1.17.0 (HEAD)
+      creationTimestamp: null
+      labels:
+        io.kompose.service: web
+      name: web
+    spec:
+      ports:
+        - name: web-endpoint
+          port: 3000
+          targetPort: 3000
+      selector:
+        io.kompose.service: web
+    status:
+      loadBalancer: { }
+  - apiVersion: v1
+    kind: Service
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert
+        kompose.version: 1.17.0 (HEAD)
+      creationTimestamp: null
+      labels:
+        io.kompose.service: elastic
+      name: elastic
+    spec:
+      ports:
+        - name: elastic-api
+          port: 9200
+          protocol: TCP
+          targetPort: 9200
+      selector:
+        io.kompose.service: elastic
+      sessionAffinity: None
+      type: ClusterIP
+  - apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert
+        kompose.version: 1.17.0 (HEAD)
+      creationTimestamp: null
+      labels:
+        io.kompose.service: elastic
+      name: elastic
+    spec:
+      replicas: 1
+      selector:
+        matchLabels:
+          io.kompose.service: elastic
+      template:
+        metadata:
+          creationTimestamp: null
+          labels:
+            io.kompose.service: elastic
+        spec:
+          containers:
+            - env:
+                - name: ES_JAVA_OPTS
+                  value: -Xms512m -Xmx512m
+                - name: discovery.type
+                  value: single-node
+              image: docker.elastic.co/elasticsearch/elasticsearch:7.10.1
+              livenessProbe:
+                exec:
+                  command:
+                    - curl --silent --fail localhost:9200/_cluster/health || exit 1
+              name: elastic
+              resources: { }
+              volumeMounts:
+                - mountPath: /usr/share/elasticsearch/data
+                  name: elastic-data
+          restartPolicy: Always
+          volumes:
+            - name: elastic-data
+              persistentVolumeClaim:
+                claimName: elastic-data
+    status: { }
+  - apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert
+        kompose.version: 1.17.0 (HEAD)
+      creationTimestamp: null
+      labels:
+        io.kompose.service: web
+      name: web
+    spec:
+      replicas: 1
+      selector:
+        matchLabels:
+          io.kompose.service: web
+      template:
+        metadata:
+          creationTimestamp: null
+          labels:
+            io.kompose.service: web
+        spec:
+          containers:
+            - env:
+                - name: REACT_APP_API_URL
+                  value: ${MONOCLE_API_URL:-http://localhost:9876}
+                - name: REACT_APP_TITLE
+                  value: ${MONOCLE_TITLE:-}
+              image: changemetrics/monocle_web:latest
+              name: web
+              ports:
+                - containerPort: 3000
+              resources: { }
+          restartPolicy: Always
+    status: { }
+  - apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      labels:
+        io.kompose.service: api
+      name: api
+    spec:
+      replicas: 1
+      selector:
+        matchLabels:
+          io.kompose.service: api
+      template:
+        metadata:
+          creationTimestamp: null
+          labels:
+            io.kompose.service: api
+        spec:
+          containers:
+            - args:
+                - uwsgi
+                - --uid
+                - guest
+                - --gid
+                - nogroup
+                - --http
+                - :9876
+                - --socket
+                - :9877
+                - --manage-script-name
+                - --mount
+                - /app=monocle.webapp:app
+              image: changemetrics/monocle_backend:latest
+              livenessProbe:
+                exec:
+                  command:
+                    - python -c "import requests,sys;
+                      r=requests.get('http://localhost:9876/api/0/health');
+                      print(r.text);
+                      sys.exit(1) if r.status_code!=200 else sys.exit(0)"
+              name: api
+              ports:
+                - containerPort: 9876
+                - containerPort: 9877
+              resources: { }
+              volumeMounts:
+                - mountPath: /etc/monocle
+                  name: config-data
+          restartPolicy: Always
+          volumes:
+            - name: config-data
+              persistentVolumeClaim:
+                claimName: config-data
+    status: { }
+  - apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert
+        kompose.version: 1.17.0 (HEAD)
+      creationTimestamp: null
+      labels:
+        io.kompose.service: crawler
+      name: crawler
+    spec:
+      replicas: 1
+      selector:
+        matchLabels:
+          io.kompose.service: crawler
+      template:
+        metadata:
+          creationTimestamp: null
+          labels:
+            io.kompose.service: crawler
+        spec:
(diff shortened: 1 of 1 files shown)


k8s/api-claim0-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  creationTimestamp: null
```
georgettica, line 4: Unnecessary, can remove :)

k8s/api-claim0-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    io.kompose.service: api-claim0
+  name: api-claim0
```
georgettica, line 7: Change the name to be more meaningful?

k8s/api-deployment.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert
+    kompose.version: 1.17.0 (HEAD)
+  creationTimestamp: null
```
georgettica, line 7: Same creation timestamp issue, maybe grep all the files for it

k8s/api-deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - name: REDIRECT_URL
+          value: ${MONOCLE_API_URL:-http://localhost:9876}/api/0/authorize
+        - name: WEB_URL
+          value: ${MONOCLE_URL:-http://localhost:3000}
+        image: changemetrics/monocle_backend:latest
+        livenessProbe:
+          exec:
+            command:
+            - python -c "import requests,sys; r=requests.get('http://localhost:9876/api/0/health');
+              print(r.text); sys.exit(1) if r.status_code!=200 else sys.exit(0)"
```
georgettica, line 55: Maybe this can be python 3 explicit?
A monocle expert can comment on this here
Also maybe split. This into multiple lines

k8s/api-deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - containerPort: 9877
+        resources: {}
+        volumeMounts:
+        - mountPath: /etc/monocle
+          name: api-claim0
+      restartPolicy: Always
+      volumes:
+      - name: api-claim0
+        persistentVolumeClaim:
+          claimName: api-claim0
```
georgettica, line 68: Claim name relates to this, change this when you change the claim name

k8s/api-deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+        ports:
+        - containerPort: 9876
+        - containerPort: 9877
+        resources: {}
+        volumeMounts:
+        - mountPath: /etc/monocle
+          name: api-claim0
+      restartPolicy: Always
+      volumes:
+      - name: api-claim0
```
georgettica, line 66: This relates to the volume mount, just fyi

georgettica (commented):
Gave some comments based on skimming the code.
We can try and deploy it to see it works, as guy said

There is one comment that needs a monocle expert to look at

georgettica:
the deployment resource needs to be changed from extensions/v1beta1 extensions/v1 because of deprecated api's

a sample deployment can be:
``` console 
$ oc create deployment test --dry-run=client -oyaml --image a:latest 
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: test
  name: test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  strategy: {}
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: test
    spec:
      containers:
      - image: a:latest
        name: a
        resources: {}
status: {}
```

k8s/api-claim0-persistentvolumeclaim.yaml
guy9050: removed:)

k8s/api-claim0-persistentvolumeclaim.yaml
guy9050: is "api-pv" okay? maybe do you have a better name?

k8s/api-deployment.yaml
guy9050: removed:)

k8s/api-deployment.yaml
guy9050: you mean to change this to python3 command?

georgettica: that can help, but I meant is splitting the command up into lines so it will be more readable

k8s/api-claim0-persistentvolumeclaim.yaml
georgettica: for example, in this pvc, the usage is in https://github.com/guy9050/monocle/blob/k8s_manifest/k8s/api-deployment.yaml#L62 and it holds the config.yaml file, but I just grepped for `api-claim0` and tried to figure out what is the actual usage of the volume

georgettica (commented):
tried to explain myself better, I hope this helps

k8s/api-claim0-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    io.kompose.service: api-claim0
+  name: api-claim0
+spec:
+  accessModes:
+  - ReadWriteOnce
```
morucci, line 9: ReadOnly ? the api process does not need write access to the volume.

k8s/api-deployment.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert
+    kompose.version: 1.17.0 (HEAD)
```
morucci, line 6: Do we need this annotation ?

k8s/api-deployment.yaml
morucci: Just python is fine here. The Alpine python-3.9 is used.
```
podman exec -it monocle_api /usr/local/bin/python --version
Python 3.9.1
```

k8s/api-deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - --gid
+        - nogroup
+        - --http
+        - :9876
+        - --socket
+        - :9877
+        - --manage-script-name
+        - --mount
+        - /app=monocle.webapp:app
+        env:
```
morucci, line 35: Would it be better to set the envs via a  configMap ?

k8s/api-claim0-persistentvolumeclaim.yaml
morucci: @georgettica +1 "config" would be the right name imo. Other deployment resources will need it too.

k8s/crawler-deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - /etc/monocle/config.yaml
+        env:
+        - name: APP_ID
+          value: ${GITHUB_APP_ID:-}
+        - name: APP_KEY_PATH
+          value: ${GITHUB_APP_KEY_PATH:-/etc/monocle/app_key.rsa}
+        image: changemetrics/monocle_backend:latest
+        name: crawler
+        resources: {}
+        volumeMounts:
```
morucci, line 38: The crawler needs the config and a RW volume to store dumps of Change that failed to be parsed.

morucci (changes requested):
Thanks that looks good ! So I've let some comments inline. Also could you add a README.md in the k8s directory to give information about how to deploy and maintain a k8s Monocle deployment ?

k8s/api-deployment.yaml
georgettica: Nope, autogenerated from kustomize and can be removed

k8s/api-deployment.yaml
georgettica: Yeah, the implementation of this was quite naive, changes can be made to separate concerns and make it more readable

k8s/api-claim0-persistentvolumeclaim.yaml
guy9050: changed

k8s/api-claim0-persistentvolumeclaim.yaml
guy9050: Changed

georgettica (commented):
after a session we had together I noticed some pointers we can fix
- naming should always be meaningfull (if we need to ask what each volume holds it's a valid request)
- maybe remove `.metadata.creationTimestamp`?
- extract to configmap (still not sure on this one but we did discuss is before)

feel free to reach out with questions @guy9050

k8s/api-deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - --gid
+        - nogroup
+        - --http
+        - :9876
+        - --socket
+        - :9877
+        - --manage-script-name
+        - --mount
+        - /app=monocle.webapp:app
+        image: changemetrics/monocle_backend:latest
```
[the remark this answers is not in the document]
morucci: Images are published automatically for each stable release (tag) but also for each master. https://github.com/change-metrics/monocle/actions/workflows/publish-master.yaml
Images are published here: https://hub.docker.com/u/changemetrics
I think this k8s manifest should use the "latest" container tag for both images and pull from dockerhub.

k8s/crawler-deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+    metadata:
+      creationTimestamp: null
+      labels:
+        io.kompose.service: crawler
+    spec:
+      containers:
+      - args:
+        - monocle
+        - --elastic-conn
+        - elastic:9200
```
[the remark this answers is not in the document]
morucci: Yes the crawler and API must be able to resolve "elastic" to elastic pod.

k8s/api-deployment.yaml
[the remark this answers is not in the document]
guy9050: can I resolve this? @morucci @georgettica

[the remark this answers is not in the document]
georgettica: We're you able to pull the image suggested? Or did you need to modify it to be image pullable..

If this needs an image change, do.the change

k8s/api-deployment.yaml
georgettica: Outdated, but the comment is still valid

k8s/crawler-deployment.yaml
georgettica: @guy9050 maybe this makes the PVC data be not only a config folder, but general days aswell

k8s/api-config-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    io.kompose.service: api-claim0
```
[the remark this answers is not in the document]
georgettica: Yes, but this change is similar across most of the pvc's

morucci (changes requested):
Could you add a README.md with instructions about how to deploy it on OpenShift ?
I'll use the instruction to test it.

Also do you think this can be validated in the GHA CI ?

k8s/api-deployment.yaml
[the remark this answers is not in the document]
guy9050: @morucci since docker has this Pull Limit rate, we are unable to pull the image. 
To workaround this we can either: sync it to Quay, or to download it into the openshift internal registry, or change pull method to be less aggressive.

k8s/api-deployment.yaml
guy9050: renamed

k8s/crawler-deployment.yaml
guy9050: @morucci there's a volume for config named "config-data" and "crawler-dump-data". Are we missing anything?

k8s/api-config-persistentvolumeclaim.yaml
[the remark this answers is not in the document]
guy9050: resolved

k8s/crawler-deployment.yaml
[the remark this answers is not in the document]
guy9050: Added the service

k8s/monocle-deployment.yml
```diff
@@ -0,0 +1,287 @@
...
+                - --elastic-conn
+                - elastic:9200
+                - crawler
+                - --config
+                - /etc/monocle/config.yaml
+              env:
+                - name: APP_ID
+                  value: ${GITHUB_APP_ID:-}
+                - name: APP_KEY_PATH
+                  value: ${GITHUB_APP_KEY_PATH:-/etc/monocle/app_key.rsa}
```
guy9050, line 270: These vars have to be kept as secrets, as k8s cannot parse them.
`Traceback (most recent call last):
  File "/usr/local/bin/monocle", line 33, in <module>
    sys.exit(load_entry_point('monocle==0.8.1', 'console_scripts', 'monocle')())
  File "/usr/local/lib/python3.9/site-packages/monocle-0.8.1-py3.9.egg/monocle/main.py", line 174, in main
  File "/usr/local/lib/python3.9/site-packages/monocle-0.8.1-py3.9.egg/monocle/github/application.py", line 201, in get_app
FileNotFoundError: [Errno 2] No such file or directory: '${GITHUB_APP_KEY_PATH:-/etc/monocle/app_key.rsa}'
`

georgettica: the meaning of the comment is that when migrating from docker-compose to k8s, this env value was kept as is

we assume there was Argument substitution and thus makes this issue not happen on docker-compose

the proposed solution is moving the env var to a secret resource and linking it into the contrainer
WDYT?

morucci: Yes moving the env into a secret resources or config map seems to be the solution as far as I understand k8s.

georgettica:
@morucci it seems you already did some k8s work using dhall.. would you want to abandon this PR in favor of your change?

morucci:
@georgettica hi, I haven't done any changes related to k8s. Which one are you talking about ?

georgettica:
I was mistakes, I saw the fhall and thought it was the k8s one


If you are already here.. can we meetup this week to talk about replicating the base config for k8s?
The config.yaml needs to be a configmap/secret and I want to put a sample resource in the deployment

morucci:
> I was mistakes, I saw the dhall and thought it was the k8s one
> 
> If you are already here.. can we meetup this week to talk about replicating the base config for k8s?
> The config.yaml needs to be a configmap/secret and I want to put a sample resource in the deployment

Sure, I'll ping you on gtalk.

PR #337, opened 2021-04-16
Improved UI

danpawlik:
Made changes:
- Small improvements in cards that includes visualizations
- Changed datepicker to be more compatible with Bootstrap
- Changed to container fluid
- Moved fitler card from top to left
- Added clear button for formular


TristanCacqueray:
The UI is being updated for patternfly, could you please have a look at the new HEAD?

PR #338, opened 2021-04-16
Rescript web

TristanCacqueray:
This change introduces the rescript language for the web interface.


TristanCacqueray:
This change also introduces a monocle Prelude.res with an abstract layout, for example the change stats pie are defined as: https://github.com/change-metrics/monocle/pull/338/commits/f9eea3167bc76713de6a14808dbeb43c993ac76f#diff-ac3e7b037e0aca7dd0f055528f411284d902bfcdd9c28c9d9a727bceb168051aR48-R57

which is implemented using re-patternfly here: https://github.com/change-metrics/monocle/pull/338/commits/f9eea3167bc76713de6a14808dbeb43c993ac76f#diff-3c425c114b2013a90d1d6595eb9e2ff665fe82bbda9e38215c3501bc59a5fd97R54-R60 

It seems like it would be simple to use another css/toolkit if we wanted to.

PR #339, opened 2021-04-18
Static Analysis CI workflow

lioramilbaum:


lioramilbaum:
@morucci Can you please review the PR?
BTW, can you please add me to the project? I would like to be able to assign myself to the issue and pick up reviewers.

Makefile
```diff
@@ -0,0 +1,8 @@
+.PHONY: dependencies static_analysis
+
+dependencies:
+	pip install --upgrade pip
```
georgettica, line 4: Shouldn't that be pip3?
As we are targeting only python3

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v1
```
georgettica, line 14: Seems there is already a v2
Did you test it before?
```suggestion
        uses: actions/setup-python@v2
```

georgettica (commented):
If you ask me, LGTM

I added some nits that can be disregarded/closed if they seem excessive

Makefile
lioramilbaum: Pushed a fix

.github/workflows/static_analysis.yml
lioramilbaum: I wasn't aware there was a new one

Omeramsc (approved):
LGTM

georgettica (approved):
Fixed all changes, full LGTM
----

If you ask me, LGTM

I added some nits that can be disregarded/closed if they seem excessive

guy9050:
LGTM

morucci:
> @morucci Can you please review the PR?
> BTW, can you please add me to the project? I would like to be able to assign myself to the issue and pick up reviewers.

@lioramilbaum I'm not familiar with configuring groups and setting ACLs with GH, after checking it more closely it does not seems obvious how to set a triage role to a group (I don't see this triage role available in the setting).

Also this https://github.blog/2019-06-25-assign-issues-to-issue-commenters/ seems a good fit. If someone want to self-assign to an issue then he needs to comment and tell me he would like to take over and fix the issue then I'll be able to assign him (he will be available in the available assignee user list). I think we can start this way and see later if we need adjustments.

tox.ini
```diff
@@ -29,7 +29,7 @@ skip_install = true
 commands =
   black --check .
   flake8 monocle
-  mypy --ignore-missing-imports monocle
+  mypy --ignore-missing-imports --python-version 3.9 monocle
```
morucci, line 32: If the plan is to make a new GHA specific to mypy then I prefer we avoid the Makefile option and instead split the tox  linters target in two new targets: format and mypy.

morucci (changes requested):
A job called "tox" seems to do already the same (defined in validate-pr.yaml). 
https://github.com/change-metrics/monocle/runs/2374648382?check_suite_focus=true#step:9:14
What is the intend of the change ?
We could maybe split the tox job in many (unit, format, mypy) ?

lioramilbaum:
> > @morucci Can you please review the PR?
> > BTW, can you please add me to the project? I would like to be able to assign myself to the issue and pick up reviewers.
> 
> @lioramilbaum I'm not familiar with configuring groups and setting ACLs with GH, after checking it more closely it does not seems obvious how to set a triage role to a group (I don't see this triage role available in the setting).
> 
> Also this https://github.blog/2019-06-25-assign-issues-to-issue-commenters/ seems a good fit. If someone want to self-assign to an issue then he needs to comment and tell me he would like to take over and fix the issue then I'll be able to assign him (he will be available in the available assignee user list). I think we can start this way and see later if we need adjustments.

We can go with whatever works for you. Can you please assign this PR?

tox.ini
lioramilbaum: Breaking tox to static_anaylisys, unit, functional can also work. The reason behind it is to make the CI quicker. All workflows are run in parallel.

lioramilbaum: > If the plan is to make a new GHA specific to mypy then I prefer we avoid the Makefile option and instead split the tox linters target in two new targets: format and mypy.

It is not a mypy specific workflow. We can split tox as you suggested.

TristanCacqueray:
I think it's good to split tests and static analysis, would you mind if I update the PR to remove the linter from the main tox job?

lioramilbaum:
> I think it's good to split tests and static analysis, would you mind if I update the PR to remove the linter from the main tox job?

Yes please. Take over the PR. I don't have to cycles to work on it.

TristanCacqueray:
@morucci i'm not able to push on this branch, the two commits of https://github.com/TristanCacqueray/monocle/commits/ci should do the trick.

TristanCacqueray (approved):
Please take a look @morucci

PR #340, opened 2021-04-19
web: use Fedora image to build the web interface

TristanCacqueray:
This change should prevent rebuilding ocaml in alpine.


web/Dockerfile
```diff
@@ -42,16 +42,16 @@ RUN npm run build
 
 ################################################################################
 
-FROM node:12-alpine
+FROM registry.fedoraproject.org/fedora:33
+
+RUN dnf update -y && dnf install -y nodejs && dnf clean all
```
morucci, line 47: We could probably remove this and move the dnf clean all line 20.

morucci (changes requested):
See inline comment please.

web/Dockerfile
TristanCacqueray: How node is going to be installed in the final image? Cleaning builder base image does not seem necessary.

morucci: Ok I missed about the multi-stage build system. Ok good so let's merge it ! thanks !

morucci (approved):

morucci:
Tested it on my laptop. It is working well. BTW we could have re-used alpine node for the production stage to avoid the nodejs installation. Anyway let's merge now !

PR #342, opened 2021-04-20
Changing Tooltip module name

danpawlik:
The rescript may raise an error on recompiling the code.
This commit is changing that.


TristanCacqueray:
The component has been moved to the re-patternfly package, but that shouldn't cause an error until we bump it to 2.0.1. Which version of re-patternfly do you have?

PR #343, opened 2021-04-22
[api] Attach task data to ChangeEvents

morucci:


monocle/webapp.py
```diff
@@ -345,6 +346,13 @@ def task_data():
                 )
             )
         total_changes_to_update = len(update_docs) - total_orphans_to_update
+        for _me in me:
+            update_docs.append(
+                TaskDataForEL(_id=_me["id"], tasks_data=mc[_me["url"]]["td"])
```
TristanCacqueray, line 351: nit: would it be possible that `me["url"]` is not part of `mc` ? e.g. in the unfortunate event that `me["url"] is http://review/1001` and only `http://review/100` is in `mc`?

Perhaps it would be safer to `try/except IndexError` this append call and log an error if that happens?

TristanCacqueray (approved):

monocle/webapp.py
morucci: Hum, yes we could add this for security but that should not happen. this "mc" is built based on the changes that match a list of url from the crawler input. Events in "me" are those that match the same list of url. We cannot have a Events w/o the corresponding change in the DB.

PR #348, opened 2021-04-22
[backend] use Fedora image to build the backend services

TristanCacqueray:
This change prevents rate-limit issue when using the docker api.


.github/workflows/validate-pr.yaml
```diff
@@ -84,11 +81,14 @@ jobs:
                     base_url: https://github.com
           EOL
 
+      - name: Start containers
+        run: docker-compose up -d
```
TristanCacqueray, line 85: Oddly enough, the container seems to start faster and they were failing because /etc/monocle/config.yaml was not available.

morucci: Yes that's not the first time I see that issue.

morucci:
I've added a new commit in the PR.
Also I've tested it locally and that seems to work as expected. I'm merging it.

PR #349, opened 2021-04-27
Replace react-scripts with esbuild

TristanCacqueray:


TristanCacqueray:
Switching to draft, it seems like this is causing unexpected issue: https://softwarefactory-project.io/paste/show/1993/

morucci (approved):

PR #351, opened 2021-04-28
[protobuf] Add initial scaffolding to manage service definition

### LINKED ISSUE #346: [spec] Use an interface description language
* Status:
* Deciders:
* Date: 2021-04-22

## Context and Problem Statement

* The current API implementations is spread across multiple modules (such as webapp, config, db and query) without a formal definition of the inputs/outputs. This induces inconsistencies and we foresee technical difficulties to maintain and evole the service.
* We would like to use an interface description language to define the API between the various component (such as between the webapi and webclients, worker and database).
* How do we define and maintain the interfaces?

## Considered Options

* OpenAPI Swagger
* Protobuf

## Decision Outcome

Chosen option: ""

### Positive Consequences

- We need to start defining interface outside of their implementations.
- We need to use a simpler HTTP api (e.g. the path for the action function name and the body for its input), instead of encoding the inputs with a mix of path, querystring and body components.

### Negative Consequences

- Follow-up decisions required for the actual implementation.

### OpenAPI

Zuul example: https://opendev.org/zuul/zuul/raw/branch/master/web/public/openapi.yaml (generated webclient: https://zuul.opendev.org/openapi ).

* Good, because it defines common HTTP APIs.
* Bad, because it is hard to maintain.

### Protobuf

Google APIs example: https://github.com/googleapis/googleapis.
ETCD APIs [model kv.proto](https://github.com/etcd-io/etcd/blob/bad0b4d5131629f0d7fe3d572cb7953548f54afd/api/mvccpb/kv.proto) and [rpc.proto](https://github.com/etcd-io/etcd/blob/bad0b4d5131629f0d7fe3d572cb7953548f54afd/api/etcdserverpb/rpc.proto).

Proof of concept for the upcoming filter sugestion api of Monocle:

```protobuf
// monocle.proto
// Copyright: (c) 2021 Monocle authors
// SPDX-License-Identifier: AGPL-3.0-only

syntax = "proto3";

// for api anotation
import "google/api/annotations.proto";

package monocle.search.v1;

option csharp_namespace = "Monocle.Search.V1";
option go_package = "github.com/change-metrics/genproto/monocle/search/v1;search";

// The messages
message GetSuggestionsRequest {
    string index = 1;
}

message GetSuggestionsResponse {
    repeated string task_types = 1;
    repeated string projects = 2;
}

// The search service
service Search {
    rpc GetSuggestions(GetSuggestionsRequest) returns (GetSuggestionsResponse) {
      option (google.api.http) = {
        post: "/api/1/get_suggestions"
        body: "*"
    };
  };
}
```

```ShellSession
# Generate openapi from protobuf definition
protoc search.proto -I/home/tdecacqu/src/github.com/googleapis/googleapis/ -I. --openapi_out=.
```

```yaml
# monocle-openapi.yaml
openapi: 3.0.3
info:
    title: Search
    description: The search service
    version: 0.0.1
paths:
    /api/1/get_suggestions:
        post:
            operationId: Search_GetSuggestions
            requestBody:
                content:
                    application/json:
                        schema:
                            $ref: '#/components/schemas/GetSuggestionsRequest'
                required: true
            responses:
                "200":
                    description: OK
                    content:
                        application/json:
                            schema:
                                $ref: '#/components/schemas/GetSuggestionsResponse'
components:
    schemas:
        GetSuggestionsRequest:
            properties:
                index:
                    type: string
            description: The messages
        GetSuggestionsResponse:
            properties:
                task_types:
                    type: array
                    items:
                        type: string
                projects:
                    type: array
                    items:
                        type: string
```

* Good, because it is a stable standard.
* Good, because it enables high quality code generation for client and server.
* Bad, because it requires custom tooling to work with.

## Links

- [gRPC vs REST: Understanding gRPC, OpenAPI and REST and when to use them in API design](https://cloud.google.com/blog/products/api-management/understanding-grpc-openapi-and-rest-and-when-to-use-them)

- [OpenAPI spec](https://spec.openapis.org/oas/v3.1.0)
- [Swagger codegen](https://github.com/swagger-api/swagger-codegen#overview)
- [Protobuf overview](https://developers.google.com/protocol-buffers/docs/overview)
- [Protobuf python tutorial](https://developers.google.com/protocol-buffers/docs/pythontutorial)
- [Protobuf javascript tutorial](https://developers.google.com/protocol-buffers/docs/reference/javascript-generated)

TristanCacqueray:
This pull request adds:

- An initial protobuf files and generate message encoder/decoder for python and rescript.
- A monocle-codegen tool to create the `webapi.py` and `WebApi.res` from the protobuf service.
- Adapt the get projects implementation to use the new types in `api.py`.

Close #346


codegen/MonocleCodegen.hs
```diff
@@ -0,0 +1,52 @@
...
+          <> map pythonAttribute fields
+      _ -> []
+    pythonAttribute :: PB.MessageField -> Text
+    pythonAttribute = \case
+      PB.NormalField _rep pbType name _num _opts -> "  " <> name <> " : " <> pythonAttributeType pbType
+      _ -> ""
+    pythonAttributeType :: PB.FieldType -> Text
+    pythonAttributeType = \case
+      PB.TString -> "str"
+      _ -> "todo"
```
morucci, line 40: Is there any existing library that handle that translate ?

TristanCacqueray: I have not found one, but that is very mechanical and it should be easy to render python or rescript type definition.

monocle/types.py
```diff
@@ -0,0 +1,28 @@
...
+from dataclasses_json import dataclass_json
+
+
+@dataclass_json
+@dataclass
+class ProjectDefinition:
+    name: str
+    repository_regex: str
+    branch_regex: str
+    file_regex: str
```
TristanCacqueray, line 16: note: protobuf does not differentiate optional and required values, and a missing attribute is set to its default value, e.g. the empty string here.

Thus the json encoded version may include the default value too, and I think that's acceptable as it makes the webclient easier to implement (because in javascript, "" is not equal to null).

We may want to make the json serialization more compact by omitting the default values in the future, but that will require some more work.

morucci (approved):

morucci:
Are you fine with this proposal @flepied  ?

TristanCacqueray:
Managing the whole protobuf json encoding by hand is too much work. So in the latest change, I've integrated  https://pypi.org/project/protobuf/ and https://www.npmjs.com/package/bs-ocaml-protoc-json to lift the json into proper data types.

monocle/webapi.py
```diff
@@ -0,0 +1,41 @@
...
+            response=json_resp, status=200, mimetype="application/json"
+        )
+
+    app.add_url_rule(
+        "/api/1/get_projects", "GetProjects", get_projects_stub, methods=["POST"]
+    )
+
+
+def task_data_service(app):
+    from monocle.api import task_data_commit
```
morucci, line 29: Maybe I miss something but I don't see it in monocle.api ?

monocle/webapp.py
```diff
@@ -50,7 +47,9 @@
     OrphanTaskDataForEL,
     TaskCrawler,
 )
+from monocle.webapi import config_service
```
morucci, line 50: task_data_service is not in use yet.

morucci (commented):
This is looking fine , that's an impressive change. It is better to use existing protobuf translator to move from defintion to python dataclasses and rescript type. I'm ok moving in this direction.

I guess here the task_data_commit endpoint should also be moved to use the new type. Then we can merge the PR.

It seems we could even generated the python client but outside of the scope of this PR.

codegen/MonocleCodegen.hs
TristanCacqueray: Actually there is one, newest changes use the protobuf python library instead.

codegen/README.md
```diff
@@ -0,0 +1,17 @@
+# Monocle codegen
+
+This tool provides a simple codegen process to create data types
+from protobuf to python and rescript.
```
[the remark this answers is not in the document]
TristanCacqueray: Heh, I made such change, it got lost in the rebase.

monocle/webapi.py
TristanCacqueray: Yes, that is not implemented yet

monocle/webapp.py
TristanCacqueray: corrent

TristanCacqueray:
Thank you for the comment. I think the remaining todos are:

- [ ] Write `doc/tutorial-protobuf.md` to describe how to create a message, encode to json and decode from json
- [ ] Create a ready to use container to deal with the protobuf compiler
- [ ] Move the task data commit code from webapp to api

codegen/MonocleCodegen.hs
```diff
@@ -0,0 +1,163 @@
+-- See https://www.schoolofhaskell.com/user/kseo/overloaded-string-literals
+{-# LANGUAGE OverloadedStrings #-}
+-- To use relude we need to disable the default standard library using NoImplicitPrelude
+{-# LANGUAGE NoImplicitPrelude #-}
+
+-- | A prototype to generate python dataclass from protobuf definition
```
morucci, line 6: No longer the case.

morucci (approved):

PR #352, opened 2021-04-29
start-with-podman.sh: Use 'z' for the relabel instead of 'Z'

morucci:


TristanCacqueray:
Perhaps we should fix the root cause instead, adding a delay seems like code smell.

TristanCacqueray (approved):

PR #354, opened 2021-04-30
Dynamically update the schema mapping and add schema version number

morucci:


morucci:
could you have a look @TristanCacqueray ?

monocle/db/db.py
```diff
@@ -244,156 +246,181 @@ def __init__(
...
+            current_index_properties = (
+                self.ic.get_mapping(index=self.index)
+                .get(self.index, {})
+                .get("mappings", {})
+                .get("properties", {})
+            )
+            for new_field_name, new_field_props in self.new_fields.items():
+                if new_field_name not in current_index_properties:
+                    body = {"properties": {new_field_name: new_field_props}}
+                    self.ic.put_mapping(index=self.index, body=body)
```
TristanCacqueray, line 414: It seems like we don't need the _schema_version field. Shouldn't this block be performed only when the version is different?

morucci: Yes let's remove the version field. We don't need it to detect if a new field mapping needs to be added. The logic is already here.

TristanCacqueray (approved):

PR #358, opened 2021-05-04
Update task data commit endpoint using protobuf

TristanCacqueray:
* [api] implement task data get last updated endpoint

* [api] implement task data commit endpoint using protobuf

* [protobuf] fix json decoding

This change uses the proper api to decode a json where a
constructed message is needed (instead of a message class).

* [protobuf] Define timestamp using Well-Known type

This change uses https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#timestamp
to define timestamp.

* [test] move linters configuration to the setup.cfg file

This change adjusts the linters test:

- run mypy first
- move configuration to a central file


monocle/tests/unit/test_api.py
```diff
@@ -369,15 +370,17 @@ def test_task_data_commit(self):
...
         # Set a new commit date and check we can retrieve it
-        input_date = "2020-01-01T01:00:00Z"
-        resp = self.client.post(posturl, json=input_date)
+        new_date = "2020-01-01T01:00:00Z"
+        postdata["timestamp"] = new_date
+        resp = self.client.post(posturl, json=postdata)
         resp = self.client.get(geturl)
         commit_date = json.loads(resp.data)
-        self.assertEqual(commit_date, input_date)
+        self.assertEqual(commit_date, new_date)
```
morucci, line 387: We could have tried a faulty request to validate to error payload too.

morucci (approved):

PR #359, opened 2021-05-05
Add filter suggestions

TristanCacqueray:


monocle/db/db.py
```diff
@@ -385,12 +385,12 @@ def __init__(
             "self_merged": {"type": "boolean"},
         }
         self.mapping.update(self.new_fields)
-        settings = {"mappings": {"properties": self.mapping}}
         self.ic = self.es.indices
         if create:
             try:
                 self.ic.get(index=self.index)
             except NotFoundError:
+                settings = {"mappings": {"properties": self.mapping}}
```
TristanCacqueray, line 393: nit: that diff does not seem related.

TristanCacqueray (commented):
LGTM

monocle/db/db.py
morucci: I'll move it.

morucci:
Does it work as expected from your side ? I no longer get Change state choice listing, neither project and approval.

TristanCacqueray:
There is indeed an issue when the filtered list is smaller than the limit: the take function from Belt.List returns a nothing. I'll fix that.

TristanCacqueray:
So a231c61 fixes the empty list issue, and  54ae78d fixes a bug where the create value was buggy.

morucci (approved):

PR #361, opened 2021-05-06
[protobuf] Add support for GET api

TristanCacqueray:
This change ensure an empty GET request can be decoded and
processed.


morucci (commented):
Some questions about it:

- is the goal is to support GET + json data as payload ?
- Should it be exposed in the http.proto and then openapi

Not related but on api/1/ endpoints a POST w/o payload data raise a 500 error. Probably a 200 with the error payload should be returned.

TristanCacqueray:
The goal is to avoid 500 error when their is no payload (for GET and POST).

PR #363, opened 2021-05-07
Change state field multi-term support

morucci:


TristanCacqueray (approved):

monocle/utils.py
```diff
@@ -241,6 +242,6 @@ def getter(attr, default):
         "task_severity",
         "task_type",
     ):
-        if params[sp]:
+        if params.get(sp):
```
TristanCacqueray, line 245: once the search endpoint is defined with protobuf, we can drop the the "," splitting business.

PR #376, opened 2021-05-24
GitLab crawler

lioramilbaum:


lioramilbaum:
@morucci I have tried to solve the broken CI with no avail. Please take a look.

monocle/gitlab/graphql.py
```diff
@@ -0,0 +1,166 @@
...
+class RequestException(Exception):
+    pass
+
+
+class GitlabGraphQLQuery(object):
+
+    log = logging.getLogger(__name__)
+
+    def __init__(self, token_getter):
+        self.url = "{github_url}/api/graphql"
```
georgettica, line 44: not sure where this is interpolated, so url name might be different
```suggestion
        self.url = "{gitlab_url}/api/graphql"
```

georgettica (commented):
@lioramilbaum  seems the CI is complaining on reformatting these files.. can you try and fix that?
```
linters run-test: commands[2] | black --force-exclude /dist-newstyle/ --check .
would reformat monocle/gitlab/graphql.py
would reformat monocle/gitlab/mergerequest.py
Oh no! 💥 💔 💥
2 files would be reformatted, 47 files would be left unchanged.
ERROR: InvocationError for command /home/runner/work/monocle/monocle/.tox/linters/bin/black --force-exclude /dist-newstyle/ --check . (exited with code 1)
```

georgettica:
didn't finish the review.. just saw your comment and responded right away

monocle/gitlab/mergerequest.py
```diff
@@ -0,0 +1,545 @@
...
+                }
+                committer {
+                  user {
+                    login
+                  }
+                }
+              }
+            }
+          }
+      """
```
georgettica, line 121: is this ok with not having a `noqa` section?

monocle/gitlab/mergerequest.py
```diff
@@ -0,0 +1,545 @@
...
+              }
+            }
+          }
+        }
+      }"""  # noqa: E501
+        data = self.gql.query(qdata % kwargs)
+        if "data" not in data:
+            self.log.error("No data collected: %s" % data)
+            if "message" in data and "wait a few minutes" in data["message"]:
+                self.log.info("sleeping 2 mn")
```
georgettica, line 264:
```suggestion
                self.log.info("sleeping 2 min")
```

monocle/gitlab/mergerequest.py
```diff
@@ -0,0 +1,545 @@
...
+                sleep(20)
+            return None
+        if not kwargs["total_prs_count"]:
+            kwargs["total_prs_count"] = data["data"]["repository"][
+                "pullRequests"
+            ]["totalCount"]
+            self.log.info(
+                "Total PRs: %s but will fetch until we reached a PR "
+                "updated at date < %s"
+                % (kwargs["total_prs_count"], kwargs["updated_since"])
```
georgettica, line 277: EDIT: seems this is the code standard, at least for this file, so feel free to ignore the comment

isn't this the old string.format method?

not judging and I am not sure if this method is planned to be deprecated (maybe it's just not in my codestyle)

monocle/gitlab/mergerequest.py
```diff
@@ -0,0 +1,545 @@
...
+            )
+            if kwargs["total_prs_count"] == 0:
+                return False
+        edges = data["data"]["repository"]["pullRequests"]["edges"]
+        for pr in edges:
+            prs.append(pr["node"])
+        oldest_update = sorted(
+            [is8601_to_dt(pr["node"]["updatedAt"]) for pr in edges],
+            reverse=True,
+        )[-1]
```
georgettica, line 287: if you are `reverse=True` why take the last one?

you could also negate it and make it clearer maybe?
```suggestion
        oldest_update = sorted(
            [is8601_to_dt(pr["node"]["updatedAt"]) for pr in edges],
        )[0]
```

monocle/gitlab/mergerequest.py
```diff
@@ -0,0 +1,545 @@
...
+            "group": group,
+            "repository": repository,
+            "number": number,
+        }
+        raw = self.gql.query(qdata % kwargs)["data"]["repository"][
+            "pullRequest"
+        ]
+        return (raw, self.extract_objects([raw], _dumper))
+
+    def extract_objects(
```
georgettica, line 378: is this not used?
if this is used please comment and I will relook at it

monocle/gitlab/mergerequest.py
```diff
@@ -0,0 +1,532 @@
...
+    )
+
+    tg = TokenGetter(args.url, args.group, args.token)
+
+    mrf = MRsFetcher(
+        graphql.GitlabGraphQLQuery(token_getter=tg),
+        args.url,
+        args.group,
+        idents_config=[],
+        repository=args.repository,
```
georgettica, line 509: why are some with `key=value` and some don't?

asking out of ignorance (and because it looks a little odd) and would like to get an answer

monocle/gitlab/graphql.py
lioramilbaum: good catch. I think url is not required. Removed it.

monocle/gitlab/mergerequest.py
```diff
@@ -0,0 +1,532 @@
...
+        pprint(prs)
+    else:
+        if not args.id:
+            print("please provide --id arg")
+            sys.exit(1)
+        data = mrf.get_one(args.url, args.group, args.repository, args.id)
+        if not args.output_dir:
+            pprint(data)
+        else:
+            basename = f"{args.url}-{args.group}-{args.repository}-{args.id}"
```
georgettica, line 525: if this is the formatting you want, just make sure it's consistent across files (I prefer this one if it helps)

requirements.txt
```diff
@@ -1,15 +1,14 @@
-requests
-elasticsearch>=7.0.0,<8.0.0
+Flask-Caching
+authlib
+cryptography<=3.3.2
+dacite==1.6.0
+elasticsearch==7.12.1
```
georgettica, line 5: is hard version pinning required or is the soft pinning not good enough (`elasticsearch>=7.0.0,<8.0.0`)

requirements.txt
```diff
@@ -1,15 +1,14 @@
-requests
-elasticsearch>=7.0.0,<8.0.0
+Flask-Caching
+authlib
+cryptography<=3.3.2
```
georgettica, line 3: removed the comment describing the issue with this module `https://github.com/pyca/cryptography/issues/5799`

if this was intentional let me know and i'll add it

monocle/gitlab/mergerequest.py
lioramilbaum: It is, as long as none of the linters and formatters complain

monocle/gitlab/mergerequest.py
lioramilbaum: It is. Updated it.

monocle/gitlab/mergerequest.py
```diff
@@ -0,0 +1,531 @@
...
+            on_author=create_ident(url, "John", []),
+            on_created_at="2020-04-11T07:00:15Z",
+            changed_files=[SimpleFile(path="README.md")],
+            approval=None,
+        )
+        objects.extend([dummy_change, dummy_event])
+        return objects
+
+
+if __name__ == "__main__":
```
morucci, line 464: I think the first step is to implement, here, in that function, an action that will fetch a single MR via a graphql query. This will help to validate how the GitLab graphql endpoint must be queried (low level details: need a token ? is there quotas system ?), but also it will help to explore the Graphql objects available to query/request a MR details.

To have that first step validated I'd like to be able to run the __main__ and give a repo + a MR id as parameter and see the MR details displayed in the console.

Second step will be to have that function able to crawl over all MRs on a given repository but since a specific date (ideally the updatedAt date). Ideally this date constraint should be passed to the GitLab grapqhql API query.

By experience, writing a crawler is difficult, so I really recommend to move forward step by step. Each API Gerrit/GitHub have there own behavior with associated bugs, so I guess it will be the same for GitLab API.

morucci (changes requested):

monocle/gitlab/mergerequest.py
lioramilbaum: I don't know yet.  This function was taken from the dummy sample.

monocle/gitlab/mergerequest.py
lioramilbaum: This was a good comment. Fixed that.

monocle/gitlab/mergerequest.py
lioramilbaum: I do not want to touch other files for this PR. I agree this formatting is preferred.

requirements.txt
lioramilbaum: hard pinning is always preferred then soft pinning or not at all. It is required for reproducible environment.

requirements.txt
lioramilbaum: A comment like that should be listed in the git commit message. That is what git is for.

monocle/gitlab/mergerequest.py
lioramilbaum: Thanks @morucci. I will go with your suggested plan.

morucci:
@lioramilbaum do you mind if I take over this change and implement the GitLab crawler ?

morucci:
The implementation of the crawler is done through the new Lentille project: https://github.com/change-metrics/lentille/pull/20

PR #377, opened 2021-05-25
[api] Initial summon of the API v2 using servant

morucci:


TristanCacqueray (approved):
Perhaps change the title to `[api] Initial summon of the API v2 using servant`

PR #392, opened 2021-06-02
codegen-haskell: Fix missing include to protobuf-src

morucci:
Error was:
sh -c 'for pb in monocle/config.proto monocle/search.proto monocle/task_data.proto; do compile-proto-file --includeDir /usr/include --includeDir protos/ --proto ${pb} --out haskell/src/; done'
Error: while processing include statements in "monocle/search.proto", failed
to find the imported file "google/protobuf/timestamp.proto", after looking in the following
locations (controlled via the --includeDir switch(es)):

  /usr/include/google/protobuf/timestamp.proto
  protos/google/protobuf/timestamp.proto

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ codegen-stubs:
 	rm -Rf srcgen/
 
 codegen-haskell:
-	sh -c 'for pb in $(MESSAGES); do compile-proto-file --includeDir /usr/include --includeDir protos/ --proto $${pb} --out haskell/src/; done'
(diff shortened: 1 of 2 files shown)


morucci:
Could you check @TristanCacqueray ?

TristanCacqueray (approved):

TristanCacqueray:
@morucci good catch, that PR make the codegen work without protobuf-devel installed globally.

PR #394, opened 2021-06-02
Changes protobuf

morucci:


Makefile
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
 MESSAGES = monocle/config.proto monocle/search.proto monocle/task_data.proto
+MESSAGES2 = monocle/change.proto
```
TristanCacqueray, line 5:
```suggestion
BACKEND_ONLY = monocle/change.proto
```

TristanCacqueray (commented):
Perhaps using a common message would be easier to manage, e.g.:

```
message Document {
  string id = 1;
  google.protobuf.Timestamp created_at = 2;
  [... common attribute]
  oneof type {
    Change change;
    ChangeCreated changeCreated;
    ChangeCommented changeCommented;
    [...]
  }
}
```

Then specific field could be provided like so:

```
message ChangeCreated {
  google.protobuf.Timestamp on_created_at = 1;
  [...]
}

message Change {
  int32 commit_count = 1;
  [...]
}
```

TristanCacqueray:
This reminds me a lot what i did for gerrit: hackage.haskell.org/package/gerrit/docs/Gerrit-Event.html :-)

morucci:
> Perhaps using a common message would be easier to manage, e.g.:
> 
> ```
> message Document {
>   string id = 1;
>   google.protobuf.Timestamp created_at = 2;
>   [... common attribute]
>   oneof type {
>     Change change;
>     ChangeCreated changeCreated;
>     ChangeCommented changeCommented;
>     [...]
>   }
> }
> ```
> 
> Then specific field could be provided like so:
> 
> ```
> message ChangeCreated {
>   google.protobuf.Timestamp on_created_at = 1;
>   [...]
> }
> 
> message Change {
>   int32 commit_count = 1;
>   [...]
> }
> ```

Yes that is a good idea I'll go with you proposal.

TristanCacqueray (approved):

PR #396, opened 2021-06-07
[web] Remove react-router and redux

TristanCacqueray:
This change replaces the legacy app design with rescript:
- Replace implicit redux with explicit reducer store
- Replace react-router with hooks

The goal is to simplify the routing login using stateless component.


TristanCacqueray:
Let's re-open when the work is actually in progress :-)

TristanCacqueray:
Unfortunately this is not feature complete, we still need to migrate the change view page!

morucci (approved):

PR #399, opened 2021-06-08
[api] add schema and use the protobuf definition of a change

morucci:


TristanCacqueray (approved):
LGTM, my only suggestion would be to name the pr `[api] add schema and use the protobuf definition of a change`

haskell/src/Monocle/Backend/Index.hs
```diff
@@ -222,6 +227,17 @@ instance ToJSON ChangesIndexMapping where
             ]
       ]
 
+createChangesIndex :: MServerName -> MIndexName -> IO (BH.BHEnv, BH.IndexName)
+createChangesIndex serverUrl indexName = do
+  let indexSettings = BH.IndexSettings (BH.ShardCount 1) (BH.ReplicaCount 0)
+  bhEnv <- Q.mkEnv serverUrl
+  let index = BH.IndexName indexName
```
TristanCacqueray: It might be safer to use the `BH.IndexName` newtype directly:
```suggestion
createChangesIndex :: MServerName -> BH.InderName -> IO (BH.BHEnv, BH.IndexName)
createChangesIndex serverUrl index = do
  let indexSettings = BH.IndexSettings (BH.ShardCount 1) (BH.ReplicaCount 0)
  bhEnv <- Q.mkEnv serverUrl
```

TristanCacqueray (approved):

PR #403, opened 2021-06-10
[index func tests] Add validation for Change doc updated

morucci:


haskell/src/Monocle/Backend/Test.hs
```diff
@@ -69,63 +70,94 @@ testIndexName :: BH.IndexName
 testIndexName = BH.IndexName "test-index"
 
 withBH :: ((BH.BHEnv, BH.IndexName) -> IO ()) -> IO ()
-withBH cb = bracket create cb delete
```
TristanCacqueray, line 72: nice ETA reduction!

TristanCacqueray (approved):

haskell/src/Monocle/Backend/Test.hs
morucci: Thanks VSCode hint ! ;)

PR #412, opened 2021-06-12
[api] Servant oauth

TristanCacqueray:
This PR implements OAuth authentication in servant and in rescript. This simplifies authentication management and this setups foundations for upcoming authenticated endpoint.


TristanCacqueray:
Let's re-open when the work is actually in progress :-)

PR #426, opened 2021-06-15
[api] Add config type through Dhall.TH

### LINKED ISSUE #407: Add `project` to the query language
The goal is to unable project based query:

* Load project configuration in servant
* Add the projects to the queryWithMods function, to resolve project expression

TristanCacqueray:
This change enables decoding the whole configuration.

Fixes #407


TristanCacqueray:
@morucci that PR would be good to have.

morucci (approved):

PR #442, opened 2021-06-29
[doc] Add host service documentation

TristanCacqueray:
This change adds documentation about running the service locally.


CONTRIBUTING.md
```diff
@@ -119,3 +119,81 @@ firefox http://localhost:8080
...
+  location / {
+    proxy_pass http://localhost:3000;
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "upgrade";
+    proxy_set_header Host $host;
+    proxy_cache_bypass $http_upgrade;
+  }
+}
+```
```
morucci, line 171: I think a script to start nginx is better. Using "podman run --net host -v $PWD/contrib/nginx/:/etc/nginx/conf.d/" to start nginx will be less intrusive to the system.

TristanCacqueray: Shouldn't we use containers to be less intrusive?

morucci (approved):

morucci:
@TristanCacqueray are you fine with my last commits on that PR ?

CONTRIBUTING.md
```diff
@@ -65,46 +67,7 @@ $ npm install
...
-
-and making it executable with `chmod +x .git/hooks/pre-commit`.
-
-## Contributing a new driver
-
-Please refer to the file `dummy/change.py` that contain a Dummy driver and
-some comments to help you get started.
-
-## Using podman to build a development environment for Monocle
+### Using podman to build a development environment for Monocle
```
TristanCacqueray, line 70: I don't think we need this, perhaps we should drop that script?

PR #445, opened 2021-07-02
apiv2 - Add organization crawler support

morucci:


haskell/src/Monocle/Backend/Index.hs
```diff
@@ -488,17 +502,28 @@ setOrUpdateLastUpdated doNotUpdate crawlerName lastUpdatedDate entity = do
...
-      otherEntity -> error $ "Unsupported Entity: " <> show otherEntity
+      Organization _ -> "organization"
 
 setLastUpdated :: Text -> UTCTime -> Entity -> TenantM ()
 setLastUpdated = setOrUpdateLastUpdated False
 
-initCrawlerLastUpdatedFromWorkerConfig :: Config.Crawler -> TenantM ()
-initCrawlerLastUpdatedFromWorkerConfig worker = traverse_ run entities
+initCrawlerEntities :: [Entity] -> Config.Crawler -> TenantM ()
+initCrawlerEntities enitities worker = traverse_ run enitities
```
TristanCacqueray, line 513:
```suggestion
initCrawlerEntities entities worker = traverse_ run entities
```

haskell/src/Monocle/Backend/Test.hs
```diff
@@ -88,13 +88,20 @@ emptyConfig name =
       index = name
    in Config.Index {..}
 
+getElasticURL :: IO Text
+getElasticURL = do
+  envM <- liftIO (lookupEnv "ELASTIC_URL")
+  pure . toText $ fromMaybe "http://localhost:9200" envM
```
TristanCacqueray, line 94: nit: no need to liftIO when the context is already `IO a`
```suggestion
getElasticURL =
  toText . fromMaybe "http://localhost:9200" <$> lookupEnv "ELASTIC_URL"
```

haskell/src/Monocle/Api/Server.hs
```diff
@@ -114,43 +115,60 @@ userGroupGet request = do
...
     Left err -> pure $ toErrorResponse err
   where
     addChanges crawlerName changes events = do
       monocleLogEvent $ AddingChange crawlerName (length changes) (length events)
       I.indexChanges (map I.toELKChange $ toList changes)
       I.indexEvents (map I.toELKChangeEvent $ toList events)
       pure $ CrawlerPB.AddDocResponse Nothing
+    addProjects crawler organizationName projects = do
+      monocleLogEvent $ AddingProject (getWorkerName crawler) organizationName (length projects)
+      I.initCrawlerEntities (Project <$> projectNames projects) crawler
```
TristanCacqueray, line 169: Why this need the `crawler` and shouldn't the function be named `I.indexProject` ?

TristanCacqueray (changes requested):
This is looking good to me, but could you check my comment on the Api.Server module?

haskell/src/Monocle/Api/Server.hs
morucci: Here we are not indexing projects but instead we set the for each project a crawler metadata entity. `setOrUpdateLastUpdated` function is used to perform this operation and need the crawler.

TristanCacqueray (approved):

PR #449, opened 2021-07-06
Config.hs: add get ident by alias

morucci:


haskell/src/Monocle/Backend/Queries.hs
```diff
@@ -146,11 +146,11 @@ getEventCounts query =
...
 
 -- $setup
--- >>> import Data.Aeson (encode)
+-- >>> import Data.Aeson (encode, decode)
 
 -- | The histo event query
--- >>> putTextLn $ decodeUtf8 $ encode histoEventAgg
--- {"date_histogram":{"interval":"day","field":"created_at"}}
+-- >>> decode $ encode histoEventAgg :: Maybe (Map Text Value)
+-- Just (fromList [("date_histogram",Object (fromList [("field",String "created_at"),("interval",String "day")]))])
```
TristanCacqueray, line 153: It seems like this key order issue remain, and using aeson-pretty may prevent that. Here is an example: https://github.com/softwarefactory-project/matrix-client-haskell/commit/4ca4963cfd06379d9bdce49742af854aed6a0d37 . Could you try something similar here, and we'll compare our result?

haskell/src/Monocle/Api/Config.hs
```diff
@@ -179,3 +179,31 @@ getTenantGroups Index {..} = Map.toList $ foldr go mempty (fromMaybe [] idents)
...
+-- >>> :{
+--  let
+--    index = emptyTenant "test" [createIdent "alice" ["opendev.org/Alice Doe/12345", "github.com/alice89"] []]
+--  in getIdentByAlias index "github.com/ghost"
+-- :}
+-- Nothing
+getIdentByAlias :: Index -> Text -> Maybe Text
+getIdentByAlias Index {..} alias = getIdentByAliasFromIdents' alias =<< idents
+  where
+    getIdentByAliasFromIdents' = flip getIdentByAliasFromIdents
```
TristanCacqueray, line 201: nit: since we introduce `getIdentByAliasFromIdents`, we could arrange its parameter to avoid the flip, then `getIdentByAlias` would be:
```suggestion
getIdentByAlias Index {..} alias = getIdentByAliasFromIdents alias =<< idents
```

haskell/src/Monocle/Api/Config.hs
```diff
@@ -179,3 +179,31 @@ getTenantGroups Index {..} = Map.toList $ foldr go mempty (fromMaybe [] idents)
...
+
+getIdentByAliasFromIdents :: [Ident] -> Text -> Maybe Text
+getIdentByAliasFromIdents idents' alias = case isMatched <$> idents' of
+  [] -> Nothing
+  x : _ -> x
+  where
+    isMatched :: Ident -> Maybe Text
+    isMatched Ident {..} = if alias `elem` aliases then Just ident else Nothing
```
TristanCacqueray, line 209: nit: the case analysis looks like the `head :: [a] -> Maybe a` from relude, so that function could be: `join $ head $ map isMatched idents'`, and if we re-arrange the parameters then we don't have to bind idents' using point free like this:

```suggestion
getIdentByAliasFromIdents :: Text -> [Ident] -> Maybe Text
getIdentByAliasFromIdents alias = join . head . map isMatched
  where
    isMatched :: Ident -> Maybe Text
    isMatched Ident {..} = if alias `elem` aliases then Just ident else Nothing
```

TristanCacqueray (changes requested):
Let's try aeson-pretty and see if that fix the doctest issue.

haskell/src/Monocle/Api/Config.hs
morucci: Here is why: https://github.com/change-metrics/lentille/pull/33/commits/3b0b7a483d35f086b6e5804e97f2bd7912975f15#diff-71dc47f1a59c491b9a150e9ad3ebdaa1a6649b745d5c27903f0fea1042da7ebeR82

haskell/src/Monocle/Backend/Queries.hs
morucci: Ok I'll see for aeson-pretty

PR #472, opened 2021-07-08
[nix] Update elasticsearch version to 7.10.1

TristanCacqueray:


morucci:
Thanks, tested with running: ELASTIC_URL=http://localhost:19200 cabal test monocle-integration-test

PR #473, opened 2021-07-08
Re-organize the web interface menu

### LINKED ISSUE #463: Re-organize the web interface menu
The goal is to improve the navigation and better organize into categories.

- [x] Rename 'Main' into 'Activity'.
- [x] Move global filter to the top menu and rename Add field to Add filter.
- [x] Move board to the change drop down.
- [x] Add a new browse project page to list the project definition.
- [x] Rename repositories into projects and move Browse repositories in dropdown, then add Browse projects.

TristanCacqueray:
Fixes #463


morucci (changes requested):
We should cover the same features than in the actual page. I'd like we add:
- Mergeablility
- Approvals
- Draft flag

Also keeping coloration might be nice to have for approvals, and merge status.

A click on the repository name or author does not refresh the page.

What do you think to move on a table view ? we could then share the component 'that display a change) from the board component.

TristanCacqueray:
I've added the missing labels, colors and fixed the author/repository link. For some reason, some change_ids are not working when clicking on the title, and I'm not sure why this is happening. Would you know why the change page query would not work?

Not sure what you mean by approvals, it was already there.

We could switch to a table view, though it's nice to see card when browsing such page.

morucci:
> I've added the missing labels, colors and fixed the author/repository link. For some reason, some change_ids are not working when clicking on the title, and I'm not sure why this is happening. Would you know why the change page query would not work?

Regarding the change_ids, the links I have in my test dataset are working fine.

Regarding the repo and author link, they update the url but the search field is not updated. I have to reload the page. Do you think we can fix this ?

> Not sure what you mean by approvals, it was already there.

I guess as my dataset is coming from the GitLab driver then I don't have the approvals ...

> We could switch to a table view, though it's nice to see card when browsing such page.

Ok I'm fine if you like it as it is.

What should we do for the graphs on that page ?

TristanCacqueray:
Ok, there may be an issue with the gerrit crawler, some change_ids are not working.

3e42191 should ensure the page reloads on click.

For the graph, I think we should move them to another page, or wait for the new query.

morucci:
Just tested it, yes I see the view is reloaded on click but it seems the browser history is not well updated. Back and Forward buttons does not react as expected. But well I think we can merge it and deal with that issue in another PR and same for the graph.

So feel free to merge :)

morucci (approved):

TristanCacqueray:
Actually we talked about the issue regarding navigation filter, e.g. when you click on a project or author, should it extend or replace the existing query?

This PR currently replaces the filter, and I think there are four ways to address that:

- Use a custom path, e.g. `/author/foo/changes`, `/project/bar/changes`, then the page would always add `author:foo` or `project:bar` to the query. But then, what should be the path for a group view timeline, e.g. `/author/foo/date/2021-07-13/changes` ?  Not very nice, otherwise the url looks nice.
- Add the filter to the existing query, e.g. if the query is `project:compute sprint42`, clicking an author would add it: `project:compute sprint42 author:foo`, but then the filter persist after the navigation.
- Use an extra query string, e.g. `f=` to pass navigation filter, then the page would add the filter to the existing query, e.g. clicking an author would result in `/changes?q=project:compute sprint42&f=author:foo`.
- Encode the navigation filters into a key value map, e.g. `a=foo&r=openstack/nova&d=2021-01-1` and then the page would convert that to a filter query.

I'll give the third options a try since it seems like the best solution, but i'm open to suggestion :-)

morucci:
> Actually we talked about the issue regarding navigation filter, e.g. when you click on a project or author, should it extend or replace the existing query?
> 
> This PR currently replaces the filter, and I think there are four ways to address that:
> 
> * Use a custom path, e.g. `/author/foo/changes`, `/project/bar/changes`, then the page would always add `author:foo` or `project:bar` to the query. But then, what should be the path for a group view timeline, e.g. `/author/foo/date/2021-07-13/changes` ?  Not very nice, otherwise the url looks nice.
> * Add the filter to the existing query, e.g. if the query is `project:compute sprint42`, clicking an author would add it: `project:compute sprint42 author:foo`, but then the filter persist after the navigation.
> * Use an extra query string, e.g. `f=` to pass navigation filter, then the page would add the filter to the existing query, e.g. clicking an author would result in `/changes?q=project:compute sprint42&f=author:foo`.

Yes this solution looks better and aligned with what is done into the board page. Each click will add a "and" to the f parameter.
+1

> * Encode the navigation filters into a key value map, e.g. `a=foo&r=openstack/nova&d=2021-01-1` and then the page would convert that to a filter query.
> 
> I'll give the third options a try since it seems like the best solution, but i'm open to suggestion :-)

TristanCacqueray:
Ok I think this is now good to go! :rocket:

morucci:
That looks good to me ! Thanks !

PR #474, opened 2021-07-09
Search order improvement

### LINKED ISSUE #464: Query improvement
The goal is to improve the query usage through the web interface. Modifiers such as 'limit' are better handled separately. This is important to enable per page query, for example to combine an author filter along with the global filter. In that situation it's tricky to support modifier in both filters.

- [x] Remove order by and limit from the query language.
- [x] Update the web search bar to include order-by and limit input.
- [x] Add field alias from and to for 'created_at<' and 'created_at>'.
- [x] Add project field.
- [x] Add search alias configuration: [(Name, Expr)].

```yaml
  search_aliases:
    - name: "approved"
      alias: "(approval:Code-Review+2)"
    - name: "sprint-01"
      alias: "from:2021-01-03 to:201-01-24"
```

TristanCacqueray:
Fixes #464

### DIFF
--- a/doc/openapi.yaml
+++ b/doc/openapi.yaml
@@ -656,6 +656,13 @@ components:
                     format: uint32
                 path:
                     type: string
+        monocle_search_Order:
+            properties:
+                field:
+                    type: string
+                direction:
+                    type: integer
+                    format: enum
         monocle_search_QueryError:
             properties:
                 message:
@@ -674,6 +681,11 @@ components:
                 query_type:
                     type: integer
                     format: enum
+                order:
+                    $ref: '#/components/schemas/monocle_search_Order'
+                limit:
+                    type: integer
+                    format: uint32
         monocle_search_QueryResponse:
             properties:
                 error:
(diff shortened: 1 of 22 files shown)


morucci (changes requested):
It seems difficult to include the display of the order setting in the board column edition. Why not just displaying it in the column or table (like it is done in the table view for the search string) ?
However, not sure if that will fit in the column view but a suggestion would be to display a tooltip when the cursor is on the column/table title to display the search string + the selected order. wdyt ?

TristanCacqueray:
Yes it is tricky to display in the editor. We can move the controller in the view below, or we could also add a line return in the editor? I was thinking the limit could also be set per column, so we need more room.

morucci:
> Yes it is tricky to display in the editor. We can move the controller in the view below

Yes it's worth a try.

Also we can probably limit the amount of keys a user can sort on.

TristanCacqueray:
We currently only support a single search order key.

@morucci could you please take a look at the latest commit?

morucci:
yes that's better. Could we remove from the order keys the one that are not displayed in the table ? It does not make sense to order on repo_regex or project for instance.

TristanCacqueray:
Ok for the regex, but why not enabling sorting per project? We could add a `sortable` attribute to the field, but in the meantime 913ff78 simply remove regex field from the controller.

morucci (approved):

PR #479, opened 2021-07-16
lentille - gitlab : Add missing event changeAbandonedEvent

morucci:


haskell/src/Lentille/GitLab/MergeRequests.hs
```diff
@@ -330,6 +331,18 @@ transformResponse host getIdentIdCB result =
...
+        getChangeAbandonedEvent :: Change -> [ChangeEvent]
+        getChangeAbandonedEvent change =
+          [ (getBaseEvent change)
+              { changeEventId = "ChangeAbandonedEvent-" <> changeId change,
+                changeEventType = Just $ ChangeEventTypeChangeAbandoned ChangeAbandonedEvent,
+                -- GitLab Graph API does not report who closed a MR
+                changeEventAuthor = changeAuthor change,
+                changeEventCreatedAt = changeUpdatedAt change
+              }
+            | isClosed $ changeState change
```
TristanCacqueray, line 343: There are no guards L317 for ChangeCreatedEvent, is that expected?

TristanCacqueray (approved):

haskell/src/Lentille/GitLab/MergeRequests.hs
morucci: That's expected. This event is always generated.

Issue #468, opened 2021-07-07
Replace the filterform with search query


morucci:
The goal is to unify the web interface using the new search query. The filterform based query needs to be rewritten using the new API.

- [x] Migrate the Activity view.
- [x] Migrate the People view.
- [x] Migrate the Project view.
- [x] Migrate the Changes view.
- [x] Investigate if Hot/Cold change is worthty to be migrated. https://github.com/change-metrics/monocle/issues/527
- [x] Remove the legacy react-router and the 'BaseQueryComponent'.

TristanCacqueray:
Project view is in progress through https://github.com/change-metrics/monocle/pull/452

PR #487, opened 2021-07-19
apiv2 - API support for top authors metrics

morucci:
related to #468

### DIFF
--- a/haskell/src/Monocle/Backend/Queries.hs
+++ b/haskell/src/Monocle/Backend/Queries.hs
@@ -300,7 +300,7 @@ getQueryFromSL query =
 getSimpleTR :: BH.TermsResult -> TermResult
 getSimpleTR tr = TermResult (getTermKey tr) (BH.termsDocCount tr)
 
-data TermResult = TermResult {term :: Text, count :: Int} deriving (Show, Eq)
+data TermResult = TermResult {tRterm :: Text, tRcount :: Int} deriving (Show, Eq)
 
 getTermKey :: BH.TermsResult -> Text
 getTermKey (BH.TermsResult (BH.TextValue tv) _ _) = tv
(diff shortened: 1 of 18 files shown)


haskell/src/Monocle/Backend/Queries.hs
```diff
@@ -300,7 +300,7 @@ getQueryFromSL query =
 getSimpleTR :: BH.TermsResult -> TermResult
 getSimpleTR tr = TermResult (getTermKey tr) (BH.termsDocCount tr)
 
-data TermResult = TermResult {term :: Text, count :: Int} deriving (Show, Eq)
+data TermResult = TermResult {tRterm :: Text, tRcount :: Int} deriving (Show, Eq)
```
TristanCacqueray, line 303: We usually keep the prefix lowercase and make the actual record field PascalCase.
```suggestion
data TermResult = TermResult {trTerm :: Text, trCount :: Int} deriving (Show, Eq)
```

TristanCacqueray: Though, why not using the schema directly from SearchPB?

TristanCacqueray (approved):

haskell/src/Monocle/Backend/Queries.hs
morucci: If you don't mind I'll prefer to keep a segmentation here

morucci:
This one is good to go

TristanCacqueray (approved):

TristanCacqueray:
Would you mind if we merge https://github.com/change-metrics/monocle/pull/496 first, then let's rebase that branch.

TristanCacqueray:
It seems like this used the wrong workflow, could we recheck?

Issue #478, opened 2021-07-14
Implement or drop support for index ACL through REMOTE_USER in apiv2


TristanCacqueray:
The legacy API enables authorized user list per index. We either need to implement this in the apiv2, or perhaps we could drop that feature since it would be better implemented with oauth or keycloak?

PR #488, opened 2021-07-19
[api] Remove index acl

TristanCacqueray:
Fixes https://github.com/change-metrics/monocle/issues/478


monocle/webapp.py
```diff
@@ -185,13 +177,7 @@ def indices():
     _indices = db.get_indices()
     indices = []
     for indice in _indices:
```
morucci, line 179: for loop is not needed

morucci (approved):

monocle/webapp.py
TristanCacqueray: let me migrate that endpoint instead.

Issue #467, opened 2021-07-07
Update documentation


morucci:
- [x] Add architecture design information.
- [ ] Add Macroscope section to configure crawlers.
- [ ] Add group definition information.

PR #489, opened 2021-07-19
Add k8s stage environment

georgettica:
Prepare resources for deployment


TristanCacqueray:
This looks good to me, though that is missing a couple of service. Let me work on https://github.com/change-metrics/monocle/issues/467 to describe the image and command.

georgettica:
I redid the k8s resource with the `kompose convert docker-compose.yaml.dev`

I'll use `kustomize` to generate a configmap from config.yaml (one for dev and one for prod)

then when done we will be able to `oc apply -k .` and it will create everything

TristanCacqueray:
Oh I see, then perhaps we should add the kompose command to the Makefile in a `codegen-kubernetes` rule?

Please have a look at https://github.com/change-metrics/monocle/pull/493 , the PR updates the compose file and features a component diagram that I hope will clear things up.

georgettica:
After some internal discussion, I have a configmap I can use and once we have the deployment set up we can discuss how to customize it to dev and prod

lioramilbaum:
@georgettica I am bit lost. Are those manifest files generated without any additional manual twick or it is 100% generated from the docker-compose file?

georgettica:
@TristanCacqueray what is the image path I should use for the deployments?

https://github.com/change-metrics/monocle/pull/489/files#diff-83c47323972de8a09244949e431e70d188d224d821062f4056a1ccb463897d8dR11-R25

I guessed it would be there

georgettica:
an option to validate the k8s resources is using https://www.kubeval.com

TristanCacqueray:
@georgettica you can also use the `docker-compose.yml.img` file, it has full the image paths.

georgettica:
i'll use that :)
didn't know that was the difference

georgettica:
@lioramilbaum this is the makefile pseudo code
https://github.com/change-metrics/monocle/pull/489/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R52-R57

georgettica:
fixed @TristanCacqueray

georgettica:
```
oc logs deploy/crawler
Missing: --monocle-url TEXT --config STRING

Usage: macroscope --monocle-url TEXT --config STRING [--debug]
                  [--interval WORD32]
```

Makefile
```diff
@@ -48,3 +48,10 @@ compose-up:
 	docker-compose up
 
 compose-restart: compose-down codegen-compose compose-up
+
+install-kustomize:
```
TristanCacqueray, line 52: The other rules of this Makefile already assume the requirements are pre-installed. I think we can drop that. You can add the tool to https://github.com/change-metrics/monocle/blob/1b7843f651fa657c56448f3dc213780b8f6c5241/nix/default.nix#L236 (e.g. `pkgs.kustomize`)

georgettica:
doesn't make sense but 
```
 oc logs deploy/web
+ test -z http://localhost:8080
+ sed -i -e s@__API_URL__@http://localhost:8080@ /app/index.html
sed: couldn't open temporary file /app/sedd6IiNH: Permission denied
+ test -z ''
+ exec nginx -g 'daemon off;'
nginx: [alert] could not open error log file: open() "/var/log/nginx/error.log" failed (13: Permission denied)
2021/07/21 13:49:40 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:5
2021/07/21 13:49:40 [emerg] 1#1: open() "/var/log/nginx/error.log" failed (13: Permission denied)
```
the volume is an emptydir... I am not sure why this is happening

Makefile
georgettica: i'll do that :)

TristanCacqueray:
> ```
> oc logs deploy/crawler
> Missing: --monocle-url TEXT --config STRING
> 
> Usage: macroscope --monocle-url TEXT --config STRING [--debug]
>                   [--interval WORD32]
> ```

This will be fixed in #493 (see: https://github.com/change-metrics/monocle/pull/493/files#diff-7e328e29c703abe09f890c3412d01a4ffdb3b7c5fedaeacf315479c0c902574d )

TristanCacqueray:
> doesn't make sense but
> 
> ```
>  oc logs deploy/web
> + test -z http://localhost:8080
> + sed -i -e s@__API_URL__@http://localhost:8080@ /app/index.html
> sed: couldn't open temporary file /app/sedd6IiNH: Permission denied
> + test -z ''
> + exec nginx -g 'daemon off;'
> nginx: [alert] could not open error log file: open() "/var/log/nginx/error.log" failed (13: Permission denied)
> 2021/07/21 13:49:40 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:5
> 2021/07/21 13:49:40 [emerg] 1#1: open() "/var/log/nginx/error.log" failed (13: Permission denied)
> ```
> 
> the volume is an emptydir... I am not sure why this is happening

I think that's an openshift related permission issue. I guess we could either:
- make the nginx /var directories o=writable
- or we could use this configuration instead: https://github.com/change-metrics/monocle/blob/1b7843f651fa657c56448f3dc213780b8f6c5241/nix/default.nix#L82-L89

TristanCacqueray:
> the volume is an emptydir... I am not sure why this is happening

Or could it be that nginx is somehow changing uid?

lioramilbaum:
@georgettica One more ask. Can you please squash the commits? Please use one commit per PR.

georgettica:
I can squash

georgettica:
@lioramilbaum can you site from where is this coming from?

Makefile
```diff
@@ -48,3 +48,11 @@ compose-up:
...
 
 compose-restart: compose-down codegen-compose compose-up
+
+# this is mentioned in nix/default.nix and should be removed at some point
+#install-kustomize:
+#	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+#	export PATH=$PATH:$(realpath .)
+
+up-stage: install-kustomize
+	k8s/build.sh | oc apply -f -
```
TristanCacqueray, line 58: We should drop the install-kustomize dependency, and we could inline the build command:

```suggestion
k8s-stage-up:
	kustomize build --load-restrictor LoadRestrictionsNone ./overlays/dev | oc apply -f -
```

k8s/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+kustomize build --load-restrictor LoadRestrictionsNone ./overlays/dev
```
TristanCacqueray, line 3: Perhaps we could drop that script, and keep such cli usage in the Makefile?

Makefile
```diff
@@ -48,3 +48,11 @@ compose-up:
 	docker-compose up
```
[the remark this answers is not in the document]
georgettica: Not a good idea, as I renamed some k8s resources and created the configmap

Feel free to do that and diff the results

nix/default.nix
```diff
@@ -233,7 +233,7 @@ let
 
   # define the base requirements
   base-req = [ pkgs.bashInteractive pkgs.coreutils pkgs.gnumake ];
-  codegen-req = [ pkgs.protobuf pkgs.ocamlPackages.ocaml-protoc pkgs.glibc ]
+  codegen-req = [ pkgs.protobuf pkgs.ocamlPackages.ocaml-protoc pkgs.glibc pkgs.kustomize ]
```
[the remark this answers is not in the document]
georgettica: Kompose is to convert to k8s from docker-compose

georgettica:
@TristanCacqueray I will leave these changes as is because @lioramilbaum had some thoughts on this

If you could help with what steps I should follow to make the error messages i sent work, I would be happy to hear

Makefile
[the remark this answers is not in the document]
TristanCacqueray: I see, well we can also patch the resources using a follow-up command. But ok, let's do that in a follow-up, perhaps using dhall directly.

nix/default.nix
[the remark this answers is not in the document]
TristanCacqueray: Alright, nevermind then.

TristanCacqueray:
> If you could help with what steps I should follow to make the error messages i sent work, I would be happy to hear

The crawler should be fixed after the rebase, if not you need to re-run the kompose process. For nginx, which solution would you prefer from the one suggested in https://github.com/change-metrics/monocle/pull/489#issuecomment-884220231 ?

georgettica:
I would say the "make nginx OpenShift compliant based"

The only thing I will say is that the base needs to be smaller or the same size as the current image

Makefile
georgettica: PS, this would need to be changed to `k8s/overlays/dev`

TristanCacqueray:
> I would say the "make nginx OpenShift compliant based"
> 
> The only thing I will say is that the base needs to be smaller or the same size as the current image

Ok, this will be done through https://github.com/change-metrics/monocle/issues/509

georgettica:
after working with @lioramilbaum on polishing the `deployment` folder we wanted to table some things we thought of:
- the PR shouldn't be blocked from merging to avoid drift between it and `master`
- a follow up PR to derive docker-compose *and* k8s yaml files  from a single source of truth based on dhall as well
- during this timeframe (from this PR being merged and the next PR merging) changes in docker-compose will need to be manually changed in the k8s resources
- additional follow up action is to also stabilize the deployment to reach a feature parity with the docker-compose counterpart

we would love to hear your opinion @TristanCacqueray @morucci on this

Makefile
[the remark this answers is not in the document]
TristanCacqueray: Looking at the proposed k8s configuration, it seems like it is currently derived from the docker-compose source. I think that is reasonable and we can maintain it separately, it should not matter much.

nix/default.nix
```diff
@@ -233,7 +233,7 @@ let
 
   # define the base requirements
   base-req = [ pkgs.bashInteractive pkgs.coreutils pkgs.gnumake ];
-  codegen-req = [ pkgs.protobuf pkgs.ocamlPackages.ocaml-protoc pkgs.glibc ]
+  codegen-req = [ pkgs.protobuf pkgs.ocamlPackages.ocaml-protoc pkgs.glibc pkgs.oc ]
```
TristanCacqueray, line 236: Unfortunately `oc` is not available in Nix, and since we are not going to codegen the k8s resources, let's remove that.

georgettica: We are going to do that in the next milestone, but I am fine with removing it for now

georgettica: Seems kubectl is available
https://nixos.wiki/wiki/Kubernetes

So let's use that!

TristanCacqueray: I meant, this list define the tool used in the Makefile.

georgettica: Yeah, and I guess a part of the deployment in nix would be to use the kubectl?

TristanCacqueray: I think it's ok to skip those requirements in nix for now, for example we don't manage docker-compose or podman here (e.g. we assume the tools are already available to the operator).

We can add those tool in nix, but let's use another list, e.g. `deploy-req = [ pkgs.kubectl pkgs.docker-compose pkgs.podman ];`

georgettica: Sure!
I just dipped into the nix thing and am excited to learn about it

I can do that

georgettica: (ps, some people still use good old docker)

lioramilbaum: I am not sure it is need to mix the nix update with this PR. Let's push it to the following one.

deployment/api-deployment.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -f ../docker-compose.yml.dev
```
morucci, line 5: why don't we use the docker-compose.yml.img as source ?

nix/default.nix
morucci: pkgs.oc can be removed.

morucci (commented):
LGTM, let's just remove the pkgs.oc from default.nix.

deployment/api-deployment.yaml
lioramilbaum: This annotation should be removed

georgettica: This was chosen as the build target was to staging and not production, and as such should be as similar to stage (dev) as possible?

I will remove these today

georgettica:
@TristanCacqueray this might be the solution we are seeking..

not sure though https://github.com/sclorg/nginx-container/blob/master/1.18/Dockerfile.fedora#L58-L75

lioramilbaum:
> @TristanCacqueray this might be the solution we are seeking..
> 
> not sure though https://github.com/sclorg/nginx-container/blob/master/1.18/Dockerfile.fedora#L58-L75

Is that solution required for this PR or the following one?

deployment/api-deployment.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.version: 1.22.0 (955b78124)
```
lioramilbaum, line 5: Please remove all references to kompose

lioramilbaum (changes requested):

deployment/api-deployment.yaml
georgettica: Though it was only that one
https://github.com/change-metrics/monocle/pull/489#discussion_r675561720

Will do

georgettica:
> > @TristanCacqueray this might be the solution we are seeking..
> > not sure though https://github.com/sclorg/nginx-container/blob/master/1.18/Dockerfile.fedora#L58-L75
> 
> Is that solution required for this PR or the following one?

the following one

deployment/api-deployment.yaml
georgettica: done :)

Makefile
georgettica: done offline

TristanCacqueray:
> > > @TristanCacqueray this might be the solution we are seeking..
> > > not sure though https://github.com/sclorg/nginx-container/blob/master/1.18/Dockerfile.fedora#L58-L75
> > 
> > 
> > Is that solution required for this PR or the following one?
> 
> the following one

I pushed the fix in https://github.com/change-metrics/monocle/pull/518

lioramilbaum:
@georgettica Is the PR ready for review? If yes, please click on the `re-request review` button for each reviewer in the list.
BTW, I can still find references to kompose in the code. Please remove all of those before request a re-review.

georgettica:
I'll do that, thought it was good enough what I did :)

And I'll re-request review

georgettica:
@lioramilbaum you can look at the latest commit to see if this looks ok by you

later tomorrow I will work on completion and merging the commit

deployment/api-deployment.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/name: api
+    app.kubernetes.io/component: api
+    app.kubernetes.io/part-of: monocl
```
morucci, line 7: Is this a typo ?

morucci (changes requested):
I see a typo, see my comment. Other than let's remove the draft mode and then we should merge and adapt with additional PR if needed.

deployment/README.md
```diff
@@ -0,0 +1,10 @@
+To deploy this run 
+
+```
+make up-stage
+```
+
+and the way this was built initially was using [kompose](https://kompose.io/)
```
lioramilbaum, line 7: Is this information valuable ?

deployment/crawler-deployment.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  creationTimestamp: null
```
lioramilbaum, line 4: please remove this key from all resources

deployment/crawler-deployment.yaml
```diff
@@ -0,0 +1,42 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      io.kompose.service: crawler
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert -f ../docker-compose.yml.dev
```
lioramilbaum, line 18: please remove kompose refferences

deployment/monocle-config.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+# created manually using the command: 
+# oc create configmap --from-file=config.yaml monocle-config --dry-run=client -oyaml
```
lioramilbaum, line 3: Is this comment valueable for understanding the code? If not, then it can be added to the commit comment or PR description.

lioramilbaum (changes requested):

deployment/monocle-config.yaml
georgettica: It is important when re-creating that resource, as the generation command currently is a bit manual

deployment/README.md
georgettica: This is important in case someone wants to re-generate, can be removed but I thought info is power

deployment/api-deployment.yaml
georgettica: It is, I'll fix that

georgettica:
@lioramilbaum  I did the change only on SOME resources (had an hour to work on it)

This is why I wanted you to look only on the yaml files I modified in the last commit

If these are good enough bi will continue with all the resources

deployment/crawler-deployment.yaml
georgettica: I will

lioramilbaum:
I suggest that we merge the PR even if it is not 100%. It would unblock me for contributing to the k8s deployment.

deployment/api-deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+              value: /etc/monocle/config.yaml
+            - name: ELASTIC_CONN
+              value: elastic:9200
+          image: quay.io/change-metrics/monocle_api:latest
+          livenessProbe:
+            exec:
+              command:
+                - python -c "import requests,sys; r=requests.post('http://localhost:9898/api/2/health', json={}); print(r.text); sys.exit(1) if r.status_code!=200 else sys.exit(0)"
+            failureThreshold: 6
+            timeoutSeconds: 60
```
TristanCacqueray, line 37: This is actually not working, the pod is in error because:

```
Warning  Unhealthy  9s                 kubelet            Liveness probe errored: rpc error: code = Unknown desc = failed to exec in container: failed to start exec "5088ef42a6771ed8fb218fd6500d1d7533d1271187e0e1ef4f010ca6333e9f52": OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "python -c \"import requests,sys; r=requests.post('http://localhost:9898/api/2/health', json={}); print(r.text); sys.exit(1) if r.status_code!=200 else sys.exit(0)\"": stat python -c "import requests,sys; r=requests.post('http://localhost:9898/api/2/health', json={}); print(r.text); sys.exit(1) if r.status_code!=200 else sys.exit(0)": no such file or directory: unknown
```

Let's remove it for now.
```suggestion
```

deployment/crawler-legacy-deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+              value: /etc/monocle/app_key.rsa
+          image: quay.io/change-metrics/monocle_backend:latest
+          name: crawler-legacy
+          resources: {}
+          volumeMounts:
+            - mountPath: /etc/monocle
+              name: monocle-config-volume
+            - mountPath: /var/lib/crawler
+              name: crawler-dump-data-volume
+      restartPolicy: Always
```
TristanCacqueray, line 44: This process actually exit when there are no crawler defined, and the pod ends up in CrashLoopBackOff status. Not sure we can fixed that here, but we should make the process loop infinitely.

TristanCacqueray (approved):
The configuration seems to work. I don't mind merging this as it is if that can help @lioramilbaum.

For the api livenessProbe, what is the best practice to simply check and http endpoint? It seems like we shouldn't need python for such a probe.

morucci (approved):

PR #492, opened 2021-07-19
apiv2 - termAgg can be parametrized for a size

morucci:
- set getRepos to 5000 by defaults.
- others most active authors functions keeping it to default and will handle size through the API request (so tackling this in another PR)


TristanCacqueray (approved):

morucci:
I did a simplification in my last push. I think this is ok to go.

TristanCacqueray:
LGTM, will merge after CI.

PR #502, opened 2021-07-21
GitLab crawler - be more resilent in case of unexpected API response format

morucci:


haskell/src/Lentille/GitLab/MergeRequests.hs
```diff
@@ -173,7 +173,11 @@ transformResponse host getIdentIdCB result =
           [],
           extract shortName fullName <$> catMaybes nodes
         )
-    otherWise -> error ("Invalid response: " <> show otherwise)
+    otherWise ->
+      ( PageInfo False Nothing Nothing,
+        ["Unknown GetProjectMergeRequests response: " <> show otherwise],
+        []
+      )
```
TristanCacqueray, line 180: The pattern match (left hand side of the -> arrow) binds a name, and in that case we already know the value, it's `result`.

Then in the term, `otherwise` is a synonym for True (unless a binding shadow it's definition, but here it is not the case because there is an upper case `W`), thus we need to show `otherWise`, or simply `result`.

```suggestion
    _anyOtherResponse ->
      ( PageInfo False Nothing Nothing,
        ["Unknown GetProjectMergeRequests response: " <> show result],
        []
      )
```

TristanCacqueray (changes requested):

haskell/src/Lentille/GitLab/MergeRequests.hs
morucci: Thanks fixed.

TristanCacqueray (approved):

PR #506, opened 2021-07-22
GHA: fix publishing workflow since container image renaming

morucci:


.github/workflows/publish-master.yaml
```diff
@@ -71,9 +71,18 @@ jobs:
...
           docker push quay.io/change-metrics/monocle_api:latest
+
+          docker tag monocle_api-legacy:latest quay.io/change-metrics/monocle_api-legacy:latest
+          docker push quay.io/change-metrics/monocle_api-legacy:latest
+
+          docker tag monocle_crawler:latest quay.io/change-metrics/monocle_crawler:latest
+          docker push quay.io/change-metrics/monocle_crawler:latest
+
+          docker tag monocle_crawler-legacy:latest quay.io/change-metrics/monocle_crawler-legacy:latest
+          docker push quay.io/change-metrics/monocle_crawler-legacy:latest
```
TristanCacqueray, line 88: We currently have monocle_api for the new api and macroscope, and monocle_backend for the legacy api and crawler. If you want to change the image, then we need to update the compose, but I think we can keep the current names.

TristanCacqueray (changes requested):

morucci:
My bad, sorry. It should be better now. Thanks for the head up !

TristanCacqueray (approved):

PR #517, opened 2021-07-24
Activity histo view

### LINKED ISSUE #468: Replace the filterform with search query

TristanCacqueray:
Fixes #468

### DIFF
--- a/haskell/src/Monocle/Backend/Queries.hs
+++ b/haskell/src/Monocle/Backend/Queries.hs
@@ -657,6 +657,42 @@ getHistoPB qf = fmap toPBHisto <$> getHisto qf
           histoCount = heCount
        in SearchPB.Histo {..}
 
+searchBody :: QueryFlavor -> Value -> QueryM Value
+searchBody qf agg = do
+  queryBH <- getQueryBHWithFlavor qf
+  pure $
+    Aeson.object
+      [ "aggregations" .= Aeson.object ["agg1" .= agg],
+        "size" .= (0 :: Word),
+        "docvalue_fields"
+          .= [ Aeson.object
+                 [ "field" .= ("created_at" :: Text),
+                   "format" .= ("date_time" :: Text)
+                 ]
+             ],
+        "query" .= fromMaybe (error "need query") queryBH
+      ]
+
+queryAggValue :: Value -> QueryM Double
+queryAggValue search = liftTenantM $ do
+  index <- getIndexName
+  res <- toAggRes <$> BHR.search index search
+  pure $ getAggValue "agg1" res
+
+averageDuration :: QueryFlavor -> QueryM Double
+averageDuration qf = queryAggValue =<< searchBody qf avg
+  where
+    avg = Aeson.object ["avg" .= Aeson.object ["field" .= ("duration" :: Text)]]
+
+medianDeviationDuration :: QueryFlavor -> QueryM Double
+medianDeviationDuration qf = queryAggValue =<< searchBody qf deviation
+  where
+    deviation =
+      Aeson.object
+        [ "median_absolute_deviation"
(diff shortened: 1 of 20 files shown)


TristanCacqueray:
- [ ] There is an issue with the changes_with_test metrics, it seems much lower than the legacy one.
- [ ] And the updates of change is higher, though that may be correct.

TristanCacqueray:
After further investigation, the previous metric were calculated using the on_created_at range. With f86c3d8 the result is mostly identical, but it seems like we need a better system to setup those flavor.

@morucci can you please take a look, I think we can merge this PR now.

haskell/src/Monocle/Backend/Queries.hs
```diff
@@ -594,8 +654,11 @@ getReviewStats = do
   let reviewStatsCommentCount = Just commentCount
       reviewStatsReviewCount = Just reviewCount
 
-  let reviewStatsCommentDelay = 0
-      reviewStatsReviewDelay = 0
+  firstComments <- withFilter [documentType ElkChangeCommentedEvent] firstEventOnChanges
+  firstReviews <- withFilter [documentType ElkChangeReviewedEvent] firstEventOnChanges
```
morucci, line 727: Nice !

haskell/src/Monocle/Backend/Queries.hs
```diff
@@ -669,10 +686,40 @@ getReviewStats = do
...
         <*> fmap countToWord (countDocs qf)
 
-    getHisto' :: ELKDocType -> QueryM (V.Vector SearchPB.Histo)
-    getHisto' docType = fmap toPBHisto <$> withFilter [mkTerm "type" (toText $ docTypeToText docType)] (getHisto qf)
-    toPBHisto :: HistoEventBucket -> SearchPB.Histo
-    toPBHisto HistoEventBucket {..} =
-      let histoDate = heDate
-          histoCount = heCount
-       in SearchPB.Histo {..}
+    getHisto' docType = withDocType docType $ getHistoPB qf
```
morucci, line 741: Awesome !

haskell/src/Monocle/Backend/Queries.hs
```diff
@@ -669,10 +686,40 @@ getReviewStats = do
...
+
+  lifecycleStatsUpdatesOfChanges <- pure 0
+  lifecycleStatsChangesWithTests <- pure 0
+  lifecycleStatsIterationsPerChange <- pure 0
+  lifecycleStatsCommitsPerChange <- pure 0
+
+  pure $ SearchPB.LifecycleStats {..}
+  where
+    qf = QueryFlavor Monocle.Search.Query.Author CreatedAt
+    getHisto' docType = withDocType docType $ getHistoPB qf
```
morucci, line 798: Could even be extracted of the where clause.

morucci (approved):
Here is two compares between master and your PR for the same dataset:

https://pasteboard.co/Kd5IdiL.png
https://pasteboard.co/Kd5IoXO.png

The metrics looks almost identical. For the change update yes, the legacy metric seems to be different and you seems to have found the reason.

- Granularity for the mean time looks different for the Reviews stats component.
- Self merged changes as well but 16 vs 17.

I don't think we should dig to much into those slightly different metric values.

I suggest to merge. And well impressive work !!!

PR #538, opened 2021-07-28
Replace inlined secrets with environment variable

### LINKED ISSUE #525: Replace inline secret with environment variable name
Instead of:

```yaml
- github_app: <token>
  github_organization: change-metrics
```

... use the environment to define secret:

```yaml
- github_app: MY_GITHUB_APP_NAME
  ...
```

TristanCacqueray:
Fixes #525

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ The process below describes how to index changes from a GitHub repository, a ful
 $ git clone https://github.com/change-metrics/monocle.git
 $ git submodule update --init --recursive
 $ cd monocle
+$ echo CRAWLERS_API_KEY=$(uuidgen) > .secrets
 $ ln -s docker-compose.yml.img docker-compose.yml
 ```
 
@@ -58,7 +59,7 @@ If you want to crawl GitHub public repositories, generate a personal access
 token on GitHub (w/o any specific rights) at https://github.com/settings/tokens.
 In case of GitHub private repositories, see the [GitHub private repositories](#github-private-repositories) section.
 
-Then create the config file `etc/config.yaml`. Here is an example your could start with. Make sure to replace `<github_token>` by your personal access token:
+Then create the config file `etc/config.yaml`. Here is an example your could start with. Make sure to write `GITHUB_TOKEN=<github_token>` in the `.secrets` file:
 
 ```YAML
 ---
@@ -67,7 +68,6 @@ workspaces:
     crawlers:
       - name: github-tektoncd
         provider:
-          github_token: <github_token>
           github_organization: tektoncd
           github_repositories:
             - operator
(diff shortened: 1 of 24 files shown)


TristanCacqueray:
@morucci that should be good for review, and it seems like we could use some cursory tests for the macroscope functions.

README.md
```diff
@@ -327,22 +317,29 @@ the initial date.
 
 ### Full configuration file example
 
+Here are the expected environment variables:
+
+- `CRAWLERS_API_KEY`: an arbitrary api key used by the crawler to index data.
+- `GITHUB_TOKEN`: a github token for `github_organization` crawler.
+- `GITLAB_TOKEN`: a gitlab token for `gitlab_organization` crawler.
+- Optional `GERRIT_PASSWORD` for `gerrit_url` crawler.
```
morucci, line 325: We should use GitHub, GitLab, Gerrit name when we talk about crawlers in the doc.

README.md
```diff
@@ -311,7 +301,7 @@ Check the OpenAPI definitions for tasks data endpoints: [Monocle OpenAPI][monocl
 ```YAML
 workspaces:
   - name: default
-    crawlers_api_key: 1a2b3c4d5e
+    crawlers_api_key: API_KEY_ENV
```
morucci, line 304: So I guess we can omit this crawlers_api_key as it is optional and then the backend is looking for CRAWLERS_API_KEY env var.

haskell/src/Monocle/Api/Config.hs
```diff
@@ -182,17 +189,19 @@ getAliases index = maybe [] (fmap toTuple) (search_aliases index)
 
 getCrawlerProject :: Crawler -> [Text]
 getCrawlerProject Crawler {..} = case provider of
-  GitlabProvider Gitlab {..} -> fromMaybe [] gitlab_repositories
+  GitlabProvider Gitlab {..} ->
+    let addOrgPrefix repo = gitlab_organization <> "/" <> repo
```
morucci, line 193: we should trim the trailing '/' for the organization name.

morucci (commented):
I'll give it a try locally.

morucci:
It seems we need to update the CONTRIBUTING.rst doc to reflect the need to have export secrets before running the apis and crawlers.

haskell/src/Monocle/Api/Server.hs
```diff
@@ -167,7 +167,7 @@ crawlerAddDoc request = do
             `orDie` CrawlerPB.AddDocErrorAddUnknownCrawler
 
         when
-          (Config.crawlers_api_key index /= toStrict apiKey)
+          (Config.crawlers_api_key index /= Just (toStrict apiKey))
```
morucci, line 170: Don't we need to read from the environment ?

morucci: I've attempted to provide a fix for this.

morucci:
Ok after my change I was able to index some gitlab and github repos.
I haven't tested the docker deployment btw.

contrib/start-crawlers-legacy.sh
```diff
@@ -0,0 +1,15 @@
...
+    python3 -mvenv .venv
+    ./.venv/bin/pip install pip
+    ./.venv/bin/pip install -r requirements.txt
+fi
+
+if ! test -f .venv/bin/monocle; then
+    ./.venv/bin/python3 setup.py install
+fi
+
+exec monocle --elastic-conn localhost:19200 crawler --config etc/config.yaml
```
morucci, line 15: Ah crap I need to set the port to 9200 by default

morucci (approved):

PR #539, opened 2021-07-28
Add prometheus client to monitor the API

TristanCacqueray:
This change adds a new middleware to serve ghc metrics on the
/metrics endpoint.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -17,6 +17,8 @@ let
   monocle-port = 19876;
   monocle2-port = 19875;
   web-port = 13000;
+  prom-port = 19090;
+  grafana-port = 19030;
 
   # DB
   info = pkgs.lib.splitString "-" pkgs.stdenv.hostPlatform.system;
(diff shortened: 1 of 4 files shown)


TristanCacqueray:
While this is already working, we should probably add the prometheus service + grafana to the compose and document its usage before merging this PR.

nix/default.nix
```diff
@@ -66,6 +68,68 @@ let
...
+  grafana-home = "/tmp/grafana-home";
+  grafanaPromDS = pkgs.writeTextFile {
+    name = "prometheus.yml";
+    text = ''
+      apiVersion: 1
+      datasources:
+      - name: Prometheus
+        type: prometheus
+        access: direct
+        url: http://localhost:19090
```
morucci, line 102: url: http://localhost:${toString prom-port}

nix/default.nix
```diff
@@ -66,6 +68,68 @@ let
...
+    text = ''
+      [security]
+      admin_user = admin
+      admin_password = monocle
+
+      [server]
+      http_port = ${toString grafana-port}
+
+      [auth.google]
+      client_secret = 0ldS3cretKey
```
morucci, line 129: Is this needed ?

morucci (approved):
I think we can merge your change. It seems to work as expected. I'll create an issue to keep track of the remaining tasks.

TristanCacqueray:
Actually the middleware might not be so useful, it would be better to manage the application metrics manually, e.g. using a label for the search endpoint for example.

morucci (approved):

PR #555, opened 2021-08-11
Add low level metrics

TristanCacqueray:


morucci (commented):
Despite the fact that it introduces quite a lot of changes in Queries.hs. I understand the intent is to ensure all queries use the same base function. However I don't see where the context is displayed in the logs.

But let's review this change together.

morucci (approved):

TristanCacqueray:
Thanks, the context is read by the `measureTenantM` function, the print is commented though.

PR #568, opened 2021-08-13
Enable profiling when context is set

TristanCacqueray:


TristanCacqueray:
Merging to fix trivial warnings

PR #569, opened 2021-08-13
Improve firstEventDuration computation by using json-syntax

### LINKED ISSUE #560: Improve query performance
The new query framework can be made more efficient:

- [x] Replace `must` by `filter` bool query
- [ ] Use the scroll/scan api instead of search

TristanCacqueray:
This change mitigates a performance issue happening with the firstEventDuration where the generic aeson decoder induce a noticable overhead.

Fixes #560

### DIFF
--- a/haskell/monocle.cabal
+++ b/haskell/monocle.cabal
@@ -222,6 +222,14 @@ executable macroscope
   import:              common-options, cli-options
   main-is:             Macroscope.hs
 
+benchmark json-decode
+  import:              common-options
+  type:                exitcode-stdio-1.0
+  hs-source-dirs:      test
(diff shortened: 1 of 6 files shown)


TristanCacqueray:
Alternatively we could investigate using cbor format for elastic response ( https://www.elastic.co/guide/en/elasticsearch/reference/7.14/sql-rest-format.html ) and/or test https://hackage.haskell.org/package/hw-json

morucci (approved):
Good catch ! We should probably evaluate a migration to json-syntax in place of Aeson. But TBH I don't see the cost of doing that in term of dev, but I bet it will improve overall perf ?

TristanCacqueray:
Unfortunately json-syntax doesn't have automatic decoder (e.g. FromJSON typeclass).

Issue #572, opened 2021-08-16
Regex limit reached for project definitions


morucci:
The current master hits this issue when a project regexps is too long. ElasticSearch limits to 1000 chars.

```
openstack: searching QueryRequest_QueryTypeQUERY_CHANGE with `(from:2021-05-01 repo_regex: .*-distgit) and project:rdo-ui`: [{"bool":{"must":[{"bool":{"must":[{"range":{"created_at":{"gt":"2021-05-01T00:00:00Z","boost":1}}},{"regexp":{"repository_fullname":{"flags":"ALL","value":".*-distgit"}}}]}},{"bool":{"must":[{"regexp":{"repository_fullname":{"flags":"ALL","value":"^rdo/openstack/octavia-dashboard-distgit|^openstack/oct
avia-dashboard|^rdo/openstack/vitrage-dashboard-distgit|^openstack/vitrage-dashboard|^rdo/openstack/horizon-distgit|^openstack/horizon|^rdo/openstack/django_openstack_auth-distgit|^openstack/django_openstack_auth|^rdo/openstack/heat-dashboard-distgit|^openstack/heat-dashboard|^rdo/openstack/ironic-ui-distgit|^openstack/ironic-ui|^rdo/openstack/cloudkitty-dashboard-distgit|^openstack/cloudkitty-dashboard|^rdo/openstack/manila
-ui-distgit|^openstack/manila-ui|^rdo/openstack/neutron-lbaas-dashboard-distgit|^openstack/neutron-lbaas-dashboard|^rdo/openstack/trove-dashboard-distgit|^openstack/trove-dashboard|^rdo/openstack/sahara-dashboard-distgit|^openstack/sahara-dashboard|^rdo/openstack/tripleo-ui-distgit|^openstack/tripleo-ui|^rdo/openstack/app-catalog-ui-distgit|^openstack/app-catalog-ui|^rdo/openstack/mistral-dashboard-distgit|^openstack/mistral
-dashboard|^rdo/openstack/magnum-ui-distgit|^openstack/magnum-ui|^rdo/openstack/murano-dashboard-distgit|^openstack/murano-dashboard|^rdo/openstack/tempest-horizon-distgit|^openstack/tempest-horizon|^rdo/openstack/karbor-dashboard-distgit|^openstack/karbor-dashboard|^rdo/openstack/designate-dashboard-distgit|^openstack/designate-dashboard"}}}]}}]}}]                                                                             
[...]

PR #577, opened 2021-08-16
Configure index with max_regex_length

TristanCacqueray:
https://github.com/change-metrics/monocle/issues/572


morucci (approved):
Thanks - tested and work as expected !

PR #586, opened 2021-08-17
Add minor api helpers

TristanCacqueray:


haskell/src/Monocle/Api/Config.hs
```diff
@@ -106,6 +106,17 @@ deriving instance Eq Index
 
 deriving instance Show Index
 
+defaultTenant :: Text -> Index
```
morucci, line 109: defaultWorkspace :) ?

morucci (approved):

PR #587, opened 2021-08-17
Add initial monoclectl deploy command

TristanCacqueray:
This change moves the nix instructions to the nix folder to
keep the contributing doc short.

This change also documents more advances usage.


TristanCacqueray:
This is not ready, i'll re-open later.

PR #595, opened 2021-08-20
Macroscope - integrate github issue crawler

morucci:


haskell/src/Lentille/GitHub/Issues.hs
```diff
@@ -140,8 +143,8 @@ transformResponse searchResult =
       where
         getLabelsE :: Either Text [LText]
         getLabelsE = case partitionEithers (getLabels issue) of
-          (labels', []) -> Right (fmap toLazy labels')
-          (_, errors) -> Left (unwords errors)
```
TristanCacqueray, line 144: Oops, that's unfortunate.

haskell/src/Monocle/Client.hs
```diff
@@ -90,3 +96,22 @@ monocleReq path MonocleClient {..} body =
     decodeResponse body' = case JSONPB.eitherDecode body' of
       Left err -> error $ "Decoding of " <> show body <> " failed with: " <> show err
       Right a -> a
+
+-- Retry 5 times network action, doubling backoff each time
+retry :: (MonadMask m, MonadLog m, MonadIO m) => m a -> m a
```
TristanCacqueray, line 101: The Prelude would have been a good place too. As long as the code doesn't depends on Monocle, i'd be happy to build reusable modules.

TristanCacqueray (approved):

PR #604, opened 2021-09-09
Migrate the Task Data API to the new APIv2

morucci:


haskell/src/Monocle/Api/Server.hs
```diff
@@ -299,6 +311,120 @@ crawlerCommitInfo request = do
...
+            . Just
+            . TaskDataPB.TaskDataGetLastUpdatedResponseResultTimestamp
+            $ T.fromUTCTime currentTS
+  where
+    toErr err =
+      TaskDataPB.TaskDataGetLastUpdatedResponse
+        . Just
+        . TaskDataPB.TaskDataGetLastUpdatedResponseResultError
+        . Enumerated
+        $ Right err
```
TristanCacqueray, line 426: nit: eta reduction

```suggestion
    toErr =
      TaskDataPB.TaskDataGetLastUpdatedResponse
        . Just
        . TaskDataPB.TaskDataGetLastUpdatedResponseResultError
        . Enumerated
        . Right
```

haskell/src/Monocle/Api/Server.hs
```diff
@@ -299,6 +311,120 @@ crawlerCommitInfo request = do
...
+      TaskDataPB.TaskDataCommitResponse
+        . Just
+        . TaskDataPB.TaskDataCommitResponseResultTimestamp
+        $ T.fromUTCTime ts'
+    toErr err =
+      TaskDataPB.TaskDataCommitResponse
+        . Just
+        . TaskDataPB.TaskDataCommitResponseResultError
+        . Enumerated
+        $ Right err
```
TristanCacqueray, line 390: nit: ditto
```suggestion
    toSuccess =
      TaskDataPB.TaskDataCommitResponse
        . Just
        . TaskDataPB.TaskDataCommitResponseResultTimestamp
        . T.fromUTCTime
    toErr =
      TaskDataPB.TaskDataCommitResponse
        . Just
        . TaskDataPB.TaskDataCommitResponseResultError
        . Enumerated
        . Right
```

haskell/src/Monocle/Api/Server.hs
```diff
@@ -299,6 +311,120 @@ crawlerCommitInfo request = do
...
+
+taskDataTaskDataCommit :: TaskDataPB.TaskDataCommitRequest -> AppM TaskDataPB.TaskDataCommitResponse
+taskDataTaskDataCommit TaskDataPB.TaskDataCommitRequest {..} = do
+  tenants <- getConfig
+  let requestE = do
+        index <-
+          Config.lookupTenant tenants (toStrict taskDataCommitRequestIndex)
+            `orDie` TaskDataPB.TaskDataCommitErrorUnknownApiKey
+
+        _crawler <-
```
TristanCacqueray, line 352:
```suggestion
        void $
```

haskell/src/Monocle/Api/Server.hs
```diff
@@ -299,6 +311,120 @@ crawlerCommitInfo request = do
...
+          Config.lookupCrawler index (toStrict taskDataCommitRequestCrawler)
+            `orDie` TaskDataPB.TaskDataCommitErrorUnknownCrawler
+        when
+          (Config.crawlers_api_key index /= Just (toStrict taskDataCommitRequestApikey))
+          (Left TaskDataPB.TaskDataCommitErrorUnknownApiKey)
+
+        void taskDataCommitRequestTimestamp `orDie` TaskDataPB.TaskDataCommitErrorCommitDateInferiorThanPrevious
+
+        pure (index, _crawler)
```
TristanCacqueray, line 361: even though we don't have an error code for that issue, we could still do the lazy validation here:
```suggestion
        let ts =  T.toUTCTime $ fromMaybe (error "Missing commit timestamp in request") taskDataCommitRequestTimestamp
        pure (index, ts)
```

haskell/src/Monocle/Backend/Index.hs
```diff
@@ -363,18 +367,46 @@ toELKChange Change {..} =
...
 
+updateDocs :: [(Value, BH.DocId)] -> TenantM ()
+updateDocs = runAddDocsBulkOPs toBulkUpdate
+  where
+    -- BulkUpdate operation: Update the document, merging the new value with the existing one.
+    toBulkUpdate index (doc, docId) = BH.BulkUpdate index docId doc
+
+-- | Generated base64 encoding of Text
+getBase64Text :: Text -> Text
+getBase64Text docId = decodeUtf8 $ B64.encode $ encodeUtf8 docId
```
TristanCacqueray, line 398: nit: eta reduction

```suggestion
getBase64Text = decodeUtf8 . B64.encode . encodeUtf8
```

protos/monocle/http.proto
```diff
@@ -91,25 +91,25 @@ service UserGroup {
 
 service TaskData {
   // Commit the last updated timestamp
-  rpc Commit(monocle_task_data.TaskDataCommitRequest) returns (monocle_task_data.TaskDataCommitResponse) {
+  rpc TaskDataCommit(monocle_task_data.TaskDataCommitRequest) returns (monocle_task_data.TaskDataCommitResponse) {
```
TristanCacqueray, line 94: Where are we now prefixing rpc and message name?

haskell/src/Monocle/Backend/Test.hs
```diff
@@ -547,6 +550,139 @@ testGetSuggestions = withTenant doTest
...
+                            tdPriority = "",
+                            tdScore = 0,
+                            tdUrl = "https://tdprovider/42-45",
+                            tdTitle = ""
+                          }
+                      ]
+                }
+            )
+        )
+        orphanTdM
```
TristanCacqueray, line 670: I guess we are missing the orphaned assignment in the new change add api? perhaps we should leave a `TODO: check adding change 45 gets the td and that it clears the orphan list.`

haskell/src/Monocle/Api/Server.hs
```diff
@@ -299,6 +311,120 @@ crawlerCommitInfo request = do
...
+
+        void $
+          Config.lookupCrawler index (toStrict taskDataAddRequestCrawler)
+            `orDie` TaskDataPB.TaskDataCommitErrorUnknownCrawler
+
+        when
+          (Config.crawlers_api_key index /= Just (toStrict taskDataAddRequestApikey))
+          (Left TaskDataPB.TaskDataCommitErrorUnknownApiKey)
+
+        pure index
```
TristanCacqueray, line 331: nit: it might be worth adding a common `validateTaskDataRequest` similar to the existing `validateSearchRequest`

TristanCacqueray (approved):
a few nits, otherwise this is LGTM. Thanks!

haskell/src/Monocle/Api/Server.hs
morucci: The binding was misnamed as I need the value later. I'll just remove the underscore.

protos/monocle/http.proto
morucci: I've got hard time to avoid conflict with Crawler's CommitRequest/CommitResponse. The solution I found was this rpc call renaming. I'm open to other solutions if any.

haskell/src/Monocle/Backend/Test.hs
morucci: Yes, I think this can be done when we will make the Gerrit and Github crawler to use the Crawler API.  At that time the  Crawler AddDocRequest function will get that capability to look if a input change match an orphan task data then will if this is the case will associate the orphan to the change and set the orphan as adopted.
No need for the TODO imo.

haskell/src/Monocle/Api/Server.hs
morucci: Sure will do.

haskell/src/Monocle/Api/Server.hs
```diff
@@ -299,6 +309,141 @@ crawlerCommitInfo request = do
...
+  -- | The index name
+  LText ->
+  -- | The crawler name
+  LText ->
+  -- | The api key
+  Maybe LText ->
+  -- | True if the timestamp must be checked
+  Bool ->
+  -- | the commit timestamp
+  Maybe T.Timestamp ->
```
TristanCacqueray, line 325: It seems like we could check the timestamp when it is set. Also it is better to avoid boolean blindness.

```suggestion
  -- | The commit timestamp
  Maybe T.Timestamp ->
```

haskell/src/Monocle/Api/Server.hs
```diff
@@ -299,6 +309,141 @@ crawlerCommitInfo request = do
...
+    crawler <- Config.lookupCrawler index (toStrict crawlerName) `orDie` TDUnknownCrawler
+    when (isJust apiKey) $
+      when (Config.crawlers_api_key index /= (toStrict <$> apiKey)) (Left TDUnknownApiKey)
+    when checkCommitDate $
+      void commitDate `orDie` TDDateInvalid
+    let ts =
+          if checkCommitDate
+            then T.toUTCTime $ fromMaybe (error "Missing commit timestamp in request") commitDate
+            else -- Probably not the better way but here return a placeholer date (will be ignored by the caller)
+              fromMaybe (error "wrong date format") (readMaybe "1980-01-01 00:00:00 Z")
```
TristanCacqueray, line 341: Thanks to lazyness, this won't blowup unless the value is accessed:

```suggestion
    let ts = T.toUTCTime $ fromMaybe (error "Missing commit timestamp in request") commitDate
```

protos/monocle/http.proto
TristanCacqueray: I see, then I guess we could improve the codegen to use qualified import (in a follow-up).

morucci:
@TristanCacqueray After some testing the change seems correct. However as of now we are using strong type I've discovered then remembered that Gerrit crawler does not set an author in case of ChangeAbandonedEvent and this make the taskData API fail when decoding events. So let's me know if we can discuss how to handle that issue as the BZ crawler (for changes from Gerrit) is unable to Add TD as the API crash.

PR #614, opened 2021-09-16
Various changes to improve UX

morucci:


morucci:
@TristanCacqueray let me know if the MonoCard is fine for your then I'll use it where that make sense.

TristanCacqueray:
@morucci that looks good, I think we can implement a `QueryRenderCard` component (with a `~title` and `~body` props) to take care of setting up the card.

TristanCacqueray (approved):

morucci:
@TristanCacqueray  ok so let's merge this and I'll followup and see how to implement this QueryRenderCard.

PR #634, opened 2021-10-01
Implement the Task Data adoption system to Haskell code base

morucci:


haskell/src/Monocle/Backend/Index.hs
```diff
@@ -422,15 +422,14 @@ upsertDocs = runAddDocsBulkOPs toBulkUpsert
 getBase64Text :: Text -> Text
 getBase64Text = decodeUtf8 . B64.encode . encodeUtf8
 
-runSimpleSearch :: FromJSON a => BH.Search -> Int -> TenantM [a]
-runSimpleSearch search size = catMaybes <$> run
+runScanSearch :: forall a. FromJSON a => BH.Search -> TenantM [a]
```
TristanCacqueray, line 425: Perhaps we should keep the `SimpleSearch` name and provides a streaming base api for proper scan search. If I understand correctly, BH.scanSearch load all the result in memory before returning the list.

TristanCacqueray (approved):

morucci:
ah already merged !

Issue #511, opened 2021-07-22
Implement a janitor service to manage ident update


TristanCacqueray:
To update ident we currently use the crawler-legacy container to run a one shot command. Instead we could use an extra service to watch for ident change and apply the difference automatically. The service could also be used to apply batch job such as curating the audit index (#453).

- [ ] Store the current config in the audit index.
- [ ] When the configuration change, applies the diff with the last one from the audit, then update the audit index.
- [ ] Remove the command from the crawler-legacy

PR #636, opened 2021-10-04
Add initial janitor toolbox

TristanCacqueray:
Related to #511


TristanCacqueray:
For simplicity, this is now squashed on #638

PR #638, opened 2021-10-04
[api] Add Streaming query scan search

TristanCacqueray:
This change implements a more efficient scanSearch using the streaming library.

### DIFF
--- a/haskell/src/Monocle/Backend/Janitor.hs
+++ b/haskell/src/Monocle/Backend/Janitor.hs
@@ -1,0 +1,28 @@
+-- | Utility function to clean up and maintain the backend
+module Monocle.Backend.Janitor (removeCrawlerData) where
+
+-- import qualified Database.Bloodhound as BH
+import qualified Monocle.Backend.Index as I
+import Monocle.Backend.Queries as Q
+import Monocle.Env
+import Monocle.Prelude
+import qualified Streaming.Prelude as Streaming
+
+-- | Remove all the data associated with a crawler name
+-- Try this on the REPL with:
+-- λ> testTenantM (mkConfig "zuul") $ removeCrawlerData "custom"
+removeCrawlerData :: Text -> TenantM ()
+removeCrawlerData crawlerName = do
+  tdDeletedCount <- removeTaskDatas
+  monocleLog $ crawlerName <> ": deleted " <> show tdDeletedCount <> " td"
+  where
+    removeTaskDatas :: TenantM Int
+    removeTaskDatas = I.bulkStream $ bulkDelete $ runQueryStream taskDataQuery Q.scanSearchId
+
+    taskDataQuery = mkQuery [mkTerm "tasks_data.crawler_name" crawlerName]
+
+    -- Helper to delete a stream of DocId
+    bulkDelete :: Stream (Of DocId) TenantM () -> Stream (Of BulkOperation) TenantM ()
+    bulkDelete s = do
+      index <- lift getIndexName
+      Streaming.map (BulkDelete index) s
--- a/haskell/src/Monocle/Backend/Queries.hs
+++ b/haskell/src/Monocle/Backend/Queries.hs
@@ -18,6 +18,8 @@ import Monocle.Prelude hiding (doSearch)
 import qualified Monocle.Search as SearchPB
 import Monocle.Search.Query (AuthorFlavor (..), QueryFlavor (..), RangeFlavor (..), rangeField)
 import qualified Monocle.Search.Query as Q
+import Streaming.Prelude (Of, Stream)
+import qualified Streaming.Prelude as Streaming
 
 -------------------------------------------------------------------------------
 -- Low level wrappers for bloodhound. Only those should be using liftTenantM.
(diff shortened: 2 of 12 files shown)


haskell/src/Monocle/Backend/Test.hs
```diff
@@ -586,16 +591,16 @@ testTaskDataAdd = withTenant doTest
           td43 = mkTaskData "43"
       void $ I.taskDataAdd [td42, td43]
       -- Ensure only changes 42 and 43 got a Task data associated
-      changes <- I.getChangesByURL (map ("https://fakeprovider/" <>) ["42", "43", "44"])
+      changes <- I.runScanSearch $ I.getChangesByURL (map ("https://fakeprovider/" <>) ["42", "43", "44"])
       assertEqual'
         "Check adding matching taskData"
-        [ ("42", Just [I.toELKTaskData td42]),
+        [ ("44", Nothing),
```
TristanCacqueray, line 597: @morucci If I understand correctly that is the correct order, e.g. the API returns the most recent first. Though I'm not sure why when we switched to the BH.scanSearch, the result somehow got reversed...

haskell/src/Monocle/Env.hs
```diff
@@ -111,6 +111,12 @@ type QueryM = ReaderT QueryEnv TenantM
 runQueryM :: Q.Query -> QueryM a -> TenantM a
 runQueryM query qm = runReaderT qm (QueryEnv query Nothing)
 
+-- | 'runQueryStream' run a QueryM stream into a TenantM
+-- Checkout the tutorial about mmoprh:
+--   https://hackage.haskell.org/package/mmorph-1.2.0/docs/Control-Monad-Morph.html#g:3
+runQueryStream :: Q.Query -> Stream (Of elem) QueryM res -> Stream (Of elem) TenantM res
+runQueryStream query = hoist (runQueryM query)
```
morucci, line 118: Could this has been implemented w/o the use of a new package ?

haskell/src/Monocle/Backend/Janitor.hs
```diff
@@ -0,0 +1,28 @@
...
+    removeTaskDatas :: TenantM Int
+    removeTaskDatas = I.bulkStream $ bulkDelete $ runQueryStream taskDataQuery Q.scanSearchId
+
+    taskDataQuery = mkQuery [mkTerm "tasks_data.crawler_name" crawlerName]
+
+    -- Helper to delete a stream of DocId
+    bulkDelete :: Stream (Of DocId) TenantM () -> Stream (Of BulkOperation) TenantM ()
+    bulkDelete s = do
+      index <- lift getIndexName
+      Streaming.map (BulkDelete index) s
```
morucci, line 28: That's cool !

morucci (approved):
That looks ok to extend the use of streaming especially for efficiency. However this adds a serious overhead around Monad business (eg. the Monad morphing lib).

haskell/src/Monocle/Backend/Test.hs
TristanCacqueray: Yes, but I wonder why #634 changed the order.

haskell/src/Monocle/Env.hs
TristanCacqueray: This is not a new package, mmorph is a dependency of streaming, see the MFunctor instance of https://hackage.haskell.org/package/streaming-0.2.3.0/docs/Streaming-Internal.html#t:Stream . If I understand correctly, `hoist` is the name of facility to operate on embedded context (e.g. when the m is not a prefix, as in `Parent m res` instead of the usual `m res`). We are doing the same thing to replace the Handler with our AppM when starting the servant server with https://hackage.haskell.org/package/servant-server-0.18.3/docs/Servant-Server.html#v:hoistServer , see https://github.com/change-metrics/monocle/blob/master/haskell/src/Monocle/Api.hs#L25.

TristanCacqueray:
I think the overhead got introduced previously with the `App <-> Tenant <-> Query` contexts where we need `runXXX` adapter to go through the layers. Perhaps this could be simplified by merging Query in Tenant, then we would not need to worry about the hoisting business.

haskell/src/Monocle/Backend/Queries.hs
```diff
@@ -59,6 +70,43 @@ doCountBH body = do
...
+    query <- getQueryBH
+    doScrollSearchBH (BHR.GetScroll "1m") (BH.mkSearch query Nothing)
+  go (getHits resp) (BH.scrollId resp)
+  where
+    -- no more result, stop here
+    go [] _ = pure ()
+    -- no more scroll, yield the result and stop
+    go xs Nothing = Streaming.each xs
+    -- otherwise, keep on scrolling
+    go xs (Just sc) = do
```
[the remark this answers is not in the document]
TristanCacqueray: It almost seems too simple to manage the scanning with streaming :-)

haskell/src/Monocle/Backend/Janitor.hs
TristanCacqueray: I think this looking great for the purpose of the Janitor.

morucci:
If both could be merged that would probably simplify the code. That might deserve a hack session on that to see if that might be possible and consequences.

PR #642, opened 2021-10-07
Consolidate the task data API with the entity API

TristanCacqueray:
This change removes the original task data API and replaces it with the new entity API.

TODO:
- [ ] Test the new macroscope stream runner.


TristanCacqueray:
The task data integration stream ran in the tests. I think this is ready to be merged.

PR #661, opened 2021-10-18
Add support for gerrit topic and hashtags

morucci:


morucci:
We could merge as is, but as you are the gerrit-haskell release owner it's up to you to tell me if you plan to the a release now or prefer to get more in before to release. Let me know.

morucci:
@TristanCacqueray well if you don't mind could you make a new release ?

Issue #664, opened 2021-10-18
Github taskdata displayed as Task: gh#I_kwDODWgvss489zoG in the web UI


morucci:
The taskdata id is used to display the label: Task: gh#I_kwDODWgvss489zoG
However the issue number should be used as taskID instead of issueID when building the data type.

PR #665, opened 2021-10-18
Fix GitHub taskdata label display #664

morucci:
The crawler is fixed to use the issue number instead of the
issue internal id.

close: #664


TristanCacqueray:
This could use a changelog update.

PR #670, opened 2021-10-21
Small code improvements related to comment in #668

morucci:
- Remove a useless use of case
- Leverage the MonadError for NoEntity case


morucci:
Thanks for the manual testing. Well I'll abandon to maybe restore later.

PR #682, opened 2021-10-26
[backend] Run the crawlers in parallel

TristanCacqueray:
This PR modifies the `Macroscope.Main` module to use a parallel execution strategy for the crawlers:

- Introduce a `MonadState Client` to group lentille client per endpoint.
- Adds a `MonadReader CrawlerEnv` constrain for LentilleM to carry the monocle client and a reference to stop the crawler when the config change.
- Implements `runCrawlers` using `UnliftIO.Async` to manage the stream group in parallel.


TristanCacqueray:
@morucci that PR seems to work as expected, it looks like a lot of changes, though the commits are smalls and the last one adds a bit of documentation I hope is useful.

Also note that this does not implement rate limit throttling yet, but we can now add threadDelay in the graphql client when needed.

haskell/src/Macroscope/Main.hs
```diff
@@ -62,7 +66,41 @@ runMacroscope confPath interval client = runLentilleM client $ do
...
+  }
+
+instance From () Clients where
+  from _ = Clients mempty mempty mempty
+
+-- | Boilerplate function to retrieve a client from the store
+getClientGerrit :: MonadGerrit m => Text -> Maybe (Text, Secret) -> StateT Clients m GerritClient
+getClientGerrit url auth = do
+  clients <- gets clientsGerrit
+  (client, newClients) <- mapMutate clients (url, auth) $ lift $ getGerritClient url auth
```
TristanCacqueray, line 172: mapMutate is introduced at the end of this commit.

TristanCacqueray (commented):
LGTM!

morucci (approved):

PR #708, opened 2021-11-02
[compose] Fix web service tests in github action and do not expose api directly

morucci:


morucci:
@TristanCacqueray I think the updated tests from that change would be valuable to test https://github.com/change-metrics/monocle/pull/700

TristanCacqueray (approved):

PR #715, opened 2021-11-05
Handle updated config schema from dhall-monocle/pull/18

morucci:


haskell/src/Monocle/Api/Config.hs
```diff
@@ -140,16 +142,14 @@ loadConfig configPath = do
   -- dhallFromYaml is able to infer the sum type by its value and it picks
   -- the first constructor that fit.
   expr <- liftIO $ Dhall.dhallFromYaml loadOpt =<< BS.readFile configPath
-  case Dhall.extract Dhall.auto expr of
-    Success config -> do
-      configWorkspaces <- traverse resolveEnv (workspaces config)
-      pure $ configWorkspaces
+  pure $ case Dhall.extract Dhall.auto expr of
+    Success config -> config
```
TristanCacqueray, line 146: nit: we could keep the resolveEnv here, e.g.:

```haskell
configWorkspaces <- traverse resolveEnv (workspaces config)
pure $ config { workspaces = configWorkspaces }
```

TristanCacqueray (approved):

haskell/src/Monocle/Api/Server.hs
```diff
@@ -27,12 +27,30 @@ import Monocle.Search.Syntax (ParseError (..))
 import qualified Monocle.UserGroup as UserGroupPB
 import Proto3.Suite (Enumerated (..))
 
+-- | /api/2/about endpoint
+configGetAbout :: ConfigPB.GetAboutRequest -> AppM ConfigPB.GetAboutResponse
+configGetAbout = const response
+  where
+    response = do
+      config <- getConfig
+      -- TODO(fbo) - fetch Monocle version from cabal file
```
TristanCacqueray, line 36: here is how i did it for podenv: https://github.com/podenv/podenv/pull/68/files# . And since we don't build the final result with nix, we can drop the `th-env` part.

haskell/src/Monocle/Env.hs
```diff
@@ -41,13 +41,17 @@ instance MonadFail AppM where
...
-getConfig :: AppM [Config.Index]
+getConfig :: AppM Config.Config
 getConfig = do
   loadConfig <- asks config
   --  logger <- asks (glLogger . aEnv)
   --  let logReload = (logEvent logger . ReloadConfig)
   liftIO (snd <$> loadConfig)
 
+-- | 'getWorkspaces' reload the workspaces automatically from the env
+getWorkspaces :: AppM [Config.Index]
```
TristanCacqueray, line 52: nit: Config.getWorkspaces does something different. It would be nice to use different name too, for example this one could be `askWorkspaces`, or the other one could be a `from` provided by `instance From Config Workspace`

TristanCacqueray (approved):
The architecture.png change does not seems relevant. Perhaps the Makefile needs to be adjusted so that it checks the file timestamps.

haskell/src/Monocle/Api/Server.hs
morucci: Thanks I'll handle it in a follow-up commit

morucci:
@TristanCacqueray  https://github.com/change-metrics/monocle/pull/722

PR #716, opened 2021-11-07
fix(web): make sed work on hardened-systems

georgettica:
based on
https://github.com/change-metrics/monocle/pull/700#issuecomment-962612490
and https://github.com/change-metrics/monocle/pull/700#discussion_r739677465

the code needs to change in order to run in hardened environments

this is the solution for it.

the `sed -i.bak` has been tested swell but it seems that even in this
state there is an intermediary step that causes this failure

### DIFF
--- a/web/conf/start.sh
+++ b/web/conf/start.sh
@@ -2,9 +2,24 @@
 # Copyright (C) 2021 Monocle authors
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+# so we don't use anything in /app that doesn't have permission
+temp_file=$(mktemp)
+# this does basically what `sed -i` does, just without writing a
+# intermedeiary file in the target folder
+# with `sed -i` the output is:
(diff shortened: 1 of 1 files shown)


TristanCacqueray (approved):
Alright, thanks!

web/conf/start.sh
```diff
@@ -2,9 +2,24 @@
...
+temp_file=$(mktemp)
+# this does basically what `sed -i` does, just without writing a
+# intermedeiary file in the target folder
+# with `sed -i` the output is:
+# "sed: couldn't open temporary file XXX: Permission denied"
+sed_temporal() {
+  local expression=$1
+  local target_file=$2
+  sed -e "${expression}" "${target_file}" > "${temp_file}"
+  cp "${temp_file}" "${target_file}"
```
morucci, line 15: you might have used mv instead of cp/rm.

georgettica: I remember mv not working for some reason, but I am not sure

Issue #598, opened 2021-08-30
:recycle: make deployment mechanism cleaner and working


georgettica:
so far the docker-compose is the de-facto way to deploy monocle

this issue will track deploying with k8s from the current state until we can bring up the whole service

Acceptance criteria:
- start with the most independent service and deploy it (will be done in a task)
- work the way up until all of the services are deployed successfully

Template for service task:
- service name
- expected issues
- possible dependencies
- add docs && diagram for service to make it clear for the user how it works and how things are interconnected

- [x] elasticsearch
- [x] api
- [x] web
- [x] crawler
- [x] legacy-crawler

georgettica:
@lioramilbaum please review whenever you see fit

georgettica:
please assign to me

georgettica:
All deoyment folder is complete, please let me know we can collaborate on #718 @TristanCacqueray @morucci

PR #717, opened 2021-11-07
fix(deploy): add liveness to all

georgettica:
now all resources have working liveness (aside legacy-crawler that I
assume will be deleted soon enough)


georgettica:
related to #598

TristanCacqueray (approved):

PR #720, opened 2021-11-08
Update start.sh to prevent failure when title contains a space

morucci:


morucci:
@georgettica could you check ? This fixes my test deployment with a TITLE set to "Monocle Devel"

georgettica:
Yeah, I will check

TristanCacqueray:
Well, with #715 , the monocle api now implements an about endpoint we could use to provide those variables, so I think we should be able to drop the start.sh soon.

georgettica:
actually, I am now getting a network error (maybe because I am exposing traffic for the `web` component and the API is not accessible on port 8080 🤷

georgettica:
I'll try that later tomorrow :)

morucci:
> Well, with #715 , the monocle api now implements an about endpoint we could use to provide those variables, so I think we should be able to drop the start.sh soon.



> Well, with #715 , the monocle api now implements an about endpoint we could use to provide those variables, so I think we should be able to drop the start.sh soon.

Probably for the title var, yes. Perhaps not for the api url. Would you like to create an issue for this ?

morucci:
I'll merge this as it fix an issue.

morucci:
@TristanCacqueray could you place you approval plz?

TristanCacqueray (approved):

PR #723, opened 2021-11-08
Display Monocle version in the About Modal

morucci:


haskell/src/Monocle/Version.hs
```diff
@@ -0,0 +1,14 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+-- | Monocle version
+module Monocle.Version (version) where
+
+import Development.GitRev (gitDirty, gitHash)
+import Relude
+
+version :: String
+version = $(gitHash) <> dirty
```
TristanCacqueray, line 10: To display the version from the cabal file, use:

```suggestion
import Data.Version (showVersion)
import Paths_monocle (version)

version :: String
version = concat [showVersion version, " ", $(gitHash), dirty]
```

And don't forget to add in the cabal file:

```
  other-modules:    Paths_monocle
  autogen-modules:  Paths_monocle
```

TristanCacqueray (approved):

haskell/src/Monocle/Version.hs
morucci: What is Paths_monocle module ?

TristanCacqueray: it is something cabal generate for us, to access the info from the cabal file.

.dockerignore
```diff
@@ -1,4 +1,3 @@
-.git
```
TristanCacqueray, line 1: I see docker build has the same issue as nix build where the .git directory is not provided. I don't think we should copy the .git folder as it will invalidate the cache.

Instead, I think we should fallback to an env variable using th-env to be set with `MONOCLE_COMMIT=$(git show HEAD --format="format:%H" -q)`

haskell/src/Monocle/Version.hs
```diff
@@ -0,0 +1,16 @@
...
+import Development.GitRev (gitDirty, gitHash)
+import qualified Paths_monocle as P (version)
+import Relude
+
+version :: String
+version = concat [showVersion P.version, " ", $(gitHash), dirty]
+  where
+    dirty
+      | $(gitDirty) = "-dirty"
+      | otherwise = ""
```
TristanCacqueray, line 16: for th-env:
```suggestion
import Language.Haskell.TH.Env (envQ)

version :: String
version = concat [showVersion P.version, " ", commit]
  where
    commit = case fromEnv of
      Just v -> v
      Nothing -> $(gitHash) <> dirty
    fromEnv :: Maybe String
    fromEnv = $$(envQ "MONOCLE_COMMIT")
    dirty
      | $(gitDirty) = "-dirty"
      | otherwise = ""
```

TristanCacqueray (changes requested):

docker-compose.dhall
```diff
@@ -119,6 +119,9 @@ let createApiService =
                         , mapValue = "/etc/monocle/config.yaml"
                         }
                       , { mapKey = "ELASTIC_CONN", mapValue = "elastic:9200" }
+                      , { mapKey = "MONOCLE_COMMIT"
+                        , mapValue = mkEnvDefault "MONOCLE_COMMIT" ""
```
TristanCacqueray, line 123: Should this be set to `HEAD` by default?

morucci:
Trying to pass as a build args instead as simple env is not working https://stackoverflow.com/questions/52429984/docker-compose-build-environment-variable

TristanCacqueray:
Should the Dockerfile-api declares a `ARG MONOCLE_COMMIT` ?

morucci:
Probably yes ...

TristanCacqueray (approved):

TristanCacqueray:
the failure does not seems to be related. I was thinking that doctest, and perhaps adding hlint / ormolu too, could be run in a secondary step, outside of cabal test suite.

morucci:
Still I'm unable to have the API about endpoint return the MONOCLE_COMMIT and I'm like out of ideas here.

TristanCacqueray (approved):

PR #729, opened 2021-11-09
[web] change hidding feature

### LINKED ISSUE #728: Add support for hidding change
When browsing the boards periodically, it might be good to be able to hide a change until it is updated:

- [ ] Add a browser local storage to store the hidden change list (and their last updated time)
- [ ] Add a `hide` button to the change widget
- [ ] Update the boards and change list to hide change (and display a reveal button when changes are hidden).

TristanCacqueray:
This PR enable the user to hide change.

Fixes #728


TristanCacqueray:
I am happy to report that dexie seems to work as expected! This implementation is a bit rough around the edge, but the feature is complete, including a setting page to clean-up the hidden change.

morucci:
Awesome ! clap
However I think that using the setting page to give ability to un-hide hidden changes is not user friendly.
I would like to propose to have a switch icon in the board and changes views called "reveal hidden changes" and when displaying changes (and reveal is set) then hidden changes appears as usual but get an un-hide icon instead of the hide icon.
We could keep the setting page code and have the icon hided for now until it serve real purpose.

morucci:
Also what about adding, for changes that re-appear after a hide, an icon that tell the user that "change got an update since your last review."

morucci:
Also please include a Changelog entry

morucci (approved):

Issue #725, opened 2021-11-08
Add GHA validation for the Kube deployment


morucci:
To protect ourself to break the deployment w/o noticing it then it might be more secure to add a GitHub action to validate the deployment.

morucci:
@georgettica would you like to tackle this issue ?

georgettica:
Sure!

georgettica:
created PR, and it seems we didn't even need to change the k8s resources (I used the strict option)

morucci:
@georgettica how could we have a GHA that run a kube deployment then verify that all pods are running ?

georgettica:
yeah, I can do that :)
but what would kind of validation would you want to run?
also, I will change in there the type of the `crawler-legacy` to Job so it doesn't run all the time

georgettica:
@morucci is there something else I can do here?
I guess the things we discussed before with the cronjob and such, but is this validation ok for now?

morucci:
Yes you are right we can close this issue. Thanks for the heads up !

PR #731, opened 2021-11-10
feat(deployment): validate k8s #725

georgettica:
you can test this locally with 'act -j check-k8s' with https://github.com/nektos/act


georgettica:
and if you want to see all of the actions, you can do that with `act -l pull_request` (because the default action is push)

morucci (approved):

morucci:
@georgettica how could we have a GHA that run a kube deployment then verify that all pods are running ?

PR #732, opened 2021-11-11
feat(actions): run k8s on minikube

georgettica:
this should be able to test locally in act -j run-k8s-deployment'

I changed all of the deployments to precede with --- to allow for 'cat'ing them together
isn't needed anymore but kept for posterity


georgettica:
this is the interesting part https://github.com/change-metrics/monocle/runs/4176450556?check_suite_focus=true

georgettica:
this is ready for review, but I guess won't pass the `kubectl wait` as elasticsearch takes a while to load

georgettica:
related to https://github.com/change-metrics/monocle/issues/725#issuecomment-965449701

morucci:
I see the error "error: no matching resources found", is it something we can help on ?

georgettica:
not really. this was a debug issue IIRC

georgettica:
this is now stuck on the fact I don't know how we should handle a tool that crashloops when it's done
I can make it a cronjob and run it during the scripts runtime.. but I don't think this is the best solution

WDYT?

morucci:
Either you could ignore crawler-legacy (do not test it) as w/o something to crawl will terminate and this is a service that we will remove (soon I hope) once the last crawler (github) is moved to crawler service. Or you can set a minimal config (even with a faulty token) to make sure the service remains up.

georgettica:
as ignoring the deployment is harder than just changing it, I migrated it to be a cronjob and this should do the trick

georgettica:
@morucci also, the haskell check keeps failing
https://github.com/change-metrics/monocle/runs/4286974052?check_suite_focus=true
can you TAL at that?

morucci:
@georgettica I see in dc22f90d70f40fd3863c9a707e777948aebf42b9 that's you update the dhall-monocle submodule ? Is that intentional ? That's maybe related.

georgettica:
Ok, so I'll revert it..
What about the docker one?

georgettica:
@morucci there is one check that is commented.. can you look at it and check if it's worth running?
Aside of that I think we are done

.github/workflows/validate-pr.yaml
```diff
@@ -129,3 +129,61 @@ jobs:
...
+        run: |
+          namespace=monocle
+          kubectl port-forward --namespace "${namespace}" $( kubectl get --namespace "${namespace}" -f deployment/elastic-service.yaml --output name) 9200:9200 &
+          sleep 5s
+
+          curl -s -I -X GET http://localhost:9200/_cat/indices
+
+          kill %1
+
+#     # this is commented as on k8s there is no nginx to route traffic
```
morucci, line 172: The web container is a nginx server that serves the web app (js) and the API via a proxy_pass option.
So you should be able to test it the same way it is done in .github/valide-pr.yaml:
```
      - name: Check api service through nginx
        run: "curl -s --fail -H 'Content-type: application/json' http://localhost:8080/api/2/get_workspaces -d '{}' | grep 'workspaces'"
```

deployment/crawler-legacy-cronjob.yaml
```diff
@@ -0,0 +1,58 @@
+---
```
morucci, line 1: That is probably ok. For the compose validation, In .github/workflows/validate-pr.yaml there is a config.yaml with a fake config to crawl a github organization. https://github.com/change-metrics/monocle/blob/master/.github/workflows/validate-pr.yaml#L65-L89 this way the crawler-legacy keeps running.

georgettica: then we should migrate the config https://github.com/change-metrics/monocle/blob/master/deployment/monocle-config.yaml to be like what you sent?

.github/workflows/validate-pr.yaml
georgettica: changed and now it works. resolving

.github/workflows/validate-pr.yaml
```diff
@@ -129,3 +129,61 @@ jobs:
...
+
+          curl -s -I -X GET http://localhost:9200/_cat/indices
+
+          kill %1
+
+      # this is commented as on k8s there is no nginx to route traffic
+      - name: Check api service through nginx
+        run: |
+          namespace=monocle
+          kubectl port-forward --namespace "${namespace}" $( kubectl get --namespace "${namespace}" -f deployment/web-service.yaml --output name) 8080:8080 &
```
morucci, line 176: Why do we need that port-forward in the CI test ? (I'm not a k8s expert :p) I mean this port must available publicly to get a working Monocle deployment. Should we include this port-forward directly into the maniftests (if that make sense) ?

.github/workflows/validate-pr.yaml
```diff
@@ -129,3 +129,61 @@ jobs:
...
+          sleep 5s
+
+          curl -s --fail -H 'Content-type: application/json' http://localhost:8080/api/2/get_workspaces -d '{}' | grep 'workspaces'
+          kill %1
+      - name: Check web service to fetch web app
+        run: |
+          namespace=monocle
+          kubectl port-forward --namespace "${namespace}" $( kubectl get --namespace "${namespace}" -f deployment/web-service.yaml --output name) 8080:8080 &
```
morucci, line 184: Same question.

.github/workflows/validate-pr.yaml
```diff
@@ -129,3 +129,61 @@ jobs:
...
+          kubectl wait --for=condition=Ready --namespace "${namespace}" pods --all
+
+          kubectl create --namespace "${namespace}" job test  --from $( kubectl get --namespace ${namespace} -f ./deployment/crawler-legacy-cronjob.yaml --output name) 
+          kubectl wait --for=condition=complete --namespace "${namespace}" job/test 
+
+      - name: Display indexes
+        run: |
+          namespace=monocle
+          kubectl port-forward --namespace "${namespace}" $( kubectl get --namespace "${namespace}" -f deployment/elastic-service.yaml --output name) 9200:9200 &
+          sleep 5s
```
morucci, line 166: nit: It seems curl have a "--retry" option

deployment/crawler-legacy-cronjob.yaml
morucci: Anyway let's not bother with this the service well be remove sooner than later.

morucci (approved):
Some nits, if you could check ?

.github/workflows/validate-pr.yaml
georgettica: So we can't, currently the deployment doesn't expose this information externally, thus port forward is required

We can add ingress to allow for external communication (which I think is only required for the webconsole?)

I think we can add this for now and I will add the ingress in another PR (as this one is taking a long time to complete)

deployment/crawler-legacy-cronjob.yaml
georgettica: So let's remove it when we obsolete it totally.. as long as we are still using it this should stay (we do run this currently)

georgettica:
Answered all the nits. I think we should keep the legacy crawler as it's still in the upstream deployment

I think this PR is working now and adding the ingress is another feature I'll need to develop on top of this

The sleep is for the `port-forward` to take form (it can also be done via checking if the port is active)

.github/workflows/validate-pr.yaml
morucci: Ok, let's do that. Any idea about the CI failure https://github.com/change-metrics/monocle/runs/4312366151?check_suite_focus=true ?

morucci: Yes only the 8080 must be exposed

.github/workflows/validate-pr.yaml
georgettica: I tried using the retry, but 404 is not recoverable (won't retry) @morucci

georgettica:
Would you want me to return the sleeps and the the code will pass?
Or is there a solution with curl that solves this?
(--retry didn't work)

morucci:
@georgettica yes please, let's go back with the sleeps then

georgettica:
Seems this is blocked until a manual approve is done :(
Tomorrow I'll squash most

morucci:
> Seems this is blocked until a manual approve is done :( Tomorrow I'll squash most

Thanks @georgettica, I'll have a final look at it on Monday and merge it manually. Do not hesitate to squash some commits if you want before.

georgettica:
@TristanCacqueray @morucci 
New mergify permissions
https://github.com/change-metrics/monocle/pull/732/checks?check_run_id=4417182506

TristanCacqueray (approved):
:+1:

georgettica:
ready to merge when PR checks are passing

morucci (approved):

.github/workflows/validate-pr.yaml
morucci: > @TristanCacqueray @morucci New mergify permissions https://github.com/change-metrics/monocle/pull/732/checks?check_run_id=4417182506

Ah seem it is : https://github.com/change-metrics/monocle/pull/773

morucci:
Mergify update  #773

PR #737, opened 2021-11-18
[nix] Update nixpkgs and bump ghc-8.10.7 and hls-1.5.0

TristanCacqueray:


TristanCacqueray:
The new shell is being built on the lambda host, then you can copy the result to your local store by running:

```
nix-copy-closure --from ssh://lambda $(nix-instantiate --attr shell ./default.nix)
```

morucci:
Thanks for the bump ! Could it be an issue to have removed all the version capping in dependencies ?
I see that doctest test is failing with "libHSmonocle-0.1.2.1-inplace-ghc8.10.7.so: invalid ELF header" any idea about the failure ?

morucci (approved):

PR #739, opened 2021-11-22
[api] Refresh index (workspace) when config is updated

### LINKED ISSUE #589: Crawler metadata not created on auto-reload of config
The API is able to reload the configuration file when changed, however crawler metadata are not refreshed thus Macroscope fails to get initial entity metadata.

morucci:
This change fixes #589.

The API keeps a state of which index received a refresh. When the `get_commit_info` endpoint is called then we verify if we need to refresh the index (looking at a TVar in the AppM Reader Env).

A refresh ensures the index is created in Elasticsearch and ensure the crawler metadata are initialized or refreshed.


haskell/src/Monocle/Api/Server.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE TupleSections #-}
```
TristanCacqueray, line 1: nit: please check https://github.com/NorfairKing/haskell-dangerous-functions#--language-tuplesections-- . Perhaps this was added because of hlint, and we should configure it to ignore that rule.

haskell/src/Monocle/Env.hs
```diff
@@ -25,7 +25,8 @@ data Env = Env
 -- | 'Env' is the global environment
 data AppEnv = AppEnv
   { config :: IO (Bool, Config.Config),
-    aEnv :: Env
+    aEnv :: Env,
+    cRStatus :: IORef [(Config.Index, Bool)]
```
TristanCacqueray, line 29: I think we need to use a TVar here because `servant` uses async handlers and we don't want to run updateCrawlerMD in parallel.

haskell/src/Monocle/Api/Server.hs
```diff
@@ -28,13 +28,65 @@ import qualified Monocle.UserGroup as UserGroupPB
...
+        $ (\ws -> (Config.getWorkspaceName ws, True)) <$> Config.getWorkspaces config
+  pure config
+
+-- | 'updateCrawlerMD' refresh crawler Metadata if needed
+updateCrawlerMD :: Config.Index -> AppM ()
+updateCrawlerMD index = do
+  shouldReload <- getCrawlerMDNeedReload
+  when shouldReload $ do
+    refreshCrawlerMD
+    setCrawlerMDReloaded
```
TristanCacqueray, line 52: It looks like this is not enough to prevent concurrent update. I think this whole function needs to happen in a `atomically . stateTVar` so that the first thread would do the initCrawlerMetadata when the bool is False, effectively blocking the other thread, which should then be able to skip the init phase.

haskell/src/Monocle/Env.hs
```diff
@@ -25,7 +25,8 @@ data Env = Env
 -- | 'Env' is the global environment
 data AppEnv = AppEnv
   { config :: IO (Bool, Config.Config),
-    aEnv :: Env
+    aEnv :: Env,
+    cRStatus :: TVar [(Text, Bool)]
```
TristanCacqueray, line 29: nit: i'm not sure what `CRStatus` means, could we find a more explicit name? Also, the `config` is using a Bool in the fst part of the tuple, should we do the same here?

Perhaps we could use a datatype to remove boolean blindness.

TristanCacqueray (approved):

PR #749, opened 2021-11-25
[ci] Run hlint static code analysis

TristanCacqueray:


morucci (approved):
Awesome no more HLint warnings \o/ (expect in codegen) !

PR #752, opened 2021-11-26
[backend] Use an unified CLI

TristanCacqueray:
This change simplifies the command API by providing an "hyper" tool
with sub commands:

- monocle-api -> monocle api
- macroscope -> monocle crawler
- lentille -> monocle lentille

Then the janitor can be implemented as another sub command.

### DIFF
--- a/deployment/api-deployment.yaml
+++ b/deployment/api-deployment.yaml
@@ -23,9 +23,8 @@ spec:
     spec:
       containers:
         - args:
-            - monocle-api
-            - --port
-            - "9898"
+            - monocle
+            - api
           env:
             - name: CONFIG
               value: /etc/monocle/config.yaml
@@ -47,7 +46,7 @@ spec:
           resources: {}
           volumeMounts:
             - mountPath: /etc/monocle
-              name: monocle-config-volume 
+              name: monocle-config-volume
       restartPolicy: Always
       volumes:
         - name: monocle-config-volume
--- a/haskell/src/CLI.hs
+++ b/haskell/src/CLI.hs
@@ -5,7 +5,8 @@
 -- | The CLI entrpoints
 module CLI (main) where
 
-import Env hiding (Parser, footer)
+import Env hiding (Parser, auto, footer)
+import qualified Env
 import qualified Lentille.Gerrit as G
 import Macroscope.Main (runMacroscope)
 import qualified Monocle.Api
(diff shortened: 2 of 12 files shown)


deployment/api-deployment.yaml
```diff
@@ -23,9 +23,8 @@ spec:
     spec:
       containers:
         - args:
-            - monocle-api
-            - --port
-            - "9898"
+            - monocle
+            - api
```
morucci, line 26: Looks like MONOCLE_API_PORT is missing from the env and we rely on the default in Monocle.CLI.

haskell/src/CLI.hs
```diff
@@ -3,80 +3,86 @@
...
+
+    -- The janitor entrypoint
+    usageJanitor = pure $ putStrLn "NotImplemented"
+
+    -- Helper to create sub command
+    mkCommand doc name parser = command name $ info parser $ progDesc doc
+
+    -- Helpers to get value fromm the env
+    getFromEnv = Env.parse (header "monocle")
+    envConf = var str "CONFIG" (help "The Monocle configuration" <> def "/etc/monocle/config.yaml" <> helpDef show)
```
morucci, line 65: I'm not able to discover the environment variables with for instance "cabal run exe:monocle -- api --help"

haskell/src/CLI.hs
```diff
@@ -3,80 +3,86 @@
...
+    envMonitoring = var str "MONOCLE_CRAWLER_MONITORING" (help "The Monitoring Port" <> def "9001")
+    getInt txt = fromMaybe (error . from $ "Invalid number: " <> txt) $ readMaybe txt
+
+main :: IO ()
+main = join $ execParser opts
+  where
+    opts =
+      info
+        (usage <**> helper)
+        (fullDesc <> progDesc "change-metrics.io | monocle")
```
morucci, line 78: changemetrics.io

haskell/src/CLI.hs
TristanCacqueray: Good idea, using https://hackage.haskell.org/package/optparse-applicative-0.16.1.0/docs/Options-Applicative.html#v:footer we can add extra doc, and we can generate the help from envparse with: https://hackage.haskell.org/package/envparse-0.4.1/docs/Env.html#v:helpDoc

morucci (approved):

morucci:
hlint: https://github.com/change-metrics/monocle/runs/4355250065?check_suite_focus=true#step:5:1382

deployment/api-deployment.yaml
TristanCacqueray: I think that's ok because this is in kubernetes internal network.

PR #755, opened 2021-11-30
[cli] Added the janitor 'update-idents' subcommand

morucci:


haskell/src/CLI.hs
```diff
@@ -84,15 +88,61 @@ getURL url =
...
+usageJanitor =
+  subparser
+    ( mkSubCommand "update-idents" "Update author identities" janitorUpdateIdent
+    )
+  where
+    janitorUpdateIdent = io <$> parser
+      where
+        parser = (,,) <$> configOption <*> elasticOption <*> workspaceOption
+        io (configPath, elasticUrl, workspaceNameM) = do
+          setEnv "CRAWLERS_API_KEY" "nokey" -- set a fake env var to resolv the configuration
```
TristanCacqueray, line 107: arg that is annoying, perhaps we need  config loader option to disable the environment lookup...

haskell/src/CLI.hs
```diff
@@ -84,15 +88,61 @@ getURL url =
...
+          case workspaceNameM of
+            Just workspaceName -> do
+              void $ case Config.lookupTenant (Config.getWorkspaces config) workspaceName of
+                Nothing -> print $ "Unable to find the workspace " <> workspaceName <> " in the Monocle config"
+                Just workspace -> runOnWorkspace env workspace
+            Nothing -> traverse_ (runOnWorkspace env) $ Config.getWorkspaces config
+        workspaceOption = optional $ strOption (long "workspace" <> O.help "Workspace name" <> metavar "WORKSPACE")
+        configOption = strOption (long "config" <> O.help "Path to configuration file" <> metavar "CONFIG")
+        elasticOption = strOption (long "elastic" <> O.help "The Elastic endpoint url" <> metavar "ELASTIC_URL")
+        runOnWorkspace env workspace = do
```
TristanCacqueray, line 119: runOnWorkspace and the body of the io action could be moved to the Janitor module too to avoid doing too much in the cli usage parser.

haskell/src/Monocle/Backend/Janitor.hs
```diff
@@ -40,6 +42,22 @@ updateIdentsOnChanges = do
   indexName <- getIndexName
   doUpdateIdentsOnChanges indexName (updateAuthor index)
 
+--- Dedicated reduced data type of EChange with only Author fields
```
TristanCacqueray, line 45: I guess you noticed performance issue with the full type. Didn't you also mention an option to also filter the fields returned by the elastic api?

TristanCacqueray (approved):

haskell/src/Monocle/Backend/Janitor.hs
morucci: I think this PR helps to improve the performance of the update-idents command thanks !
https://github.com/change-metrics/monocle/pull/757

morucci:
Ok I've validated the compose command for update-idents on a compose deployment and that works as expected. Let's merge the change then if last commit is fine for you @TristanCacqueray

TristanCacqueray (approved):

PR #756, opened 2021-11-30
[crawler] graphQL - fix recently introduced issue where cursor is not…

morucci:
… provided

Fix error at crawler startup: "Missing endCursor".

### DIFF
--- a/haskell/src/Lentille/GraphQL.hs
+++ b/haskell/src/Lentille/GraphQL.hs
@@ -110,7 +110,7 @@ streamFetch client mkArgs transformResponse = go Nothing
         lift $
           fetchWithLog
             (doGraphRequest client)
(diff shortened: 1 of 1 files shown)


haskell/src/Lentille/GraphQL.hs
```diff
@@ -110,7 +110,7 @@ streamFetch client mkArgs transformResponse = go Nothing
         lift $
           fetchWithLog
             (doGraphRequest client)
-            (mkArgs . fromMaybe (error "Missing endCursor") $ (Just . endCursor) =<< pageInfoM)
+            (mkArgs $ endCursor =<< pageInfoM)
```
TristanCacqueray, line 113: We might want to still fail if the endCursor is missing from the pageInfo, e.g.:

```suggestion
            (mkArgs $ (Just . fromMaybe (error "Missing endCursor from page info") . endCursor) =<< pageInfoM)
```

TristanCacqueray (approved):

PR #765, opened 2021-12-01
Add support for username configuration to use `self` in queries

TristanCacqueray:
This PR let a user provide a username through the url:

- `&username=alice` : use this name and save it in the local storage
- `&username=` : remove the name from the local storage
- ``: use the name from the local storage, if any

This could use a better interface through some sort of a login modal, and to provide an indicator about the login state.


morucci (approved):
I've tested and that work as expected.

morucci:
However as you said it might be better to get a better UX for setting the login. For instance I got trouble with a fragment in the URL as I put "&username=" after it ...

I think we can merge your change and work on a better UX in a followup change.

PR #771, opened 2021-12-03
[web] Add login validation

morucci:


haskell/src/Monocle/Api/Server.hs
```diff
@@ -62,6 +63,30 @@ pattern GetConfig a <- Config.ConfigStatus _ a _
 pattern GetTenants :: [Config.Index] -> Config.ConfigStatus
 pattern GetTenants a <- Config.ConfigStatus _ (Config.Config _ a) _
 
+-- | /login/validate endpoint
+loginLoginValidation ::
+  LoginPB.LoginValidationRequest -> AppM LoginPB.LoginValidationResponse
+loginLoginValidation request = do
+  GetTenants tenants <- getConfig
+  let username = request & LoginPB.loginValidationRequestUsername
+  validated <- traverse (validateOnIndex $ from username) tenants
```
TristanCacqueray, line 72: I published a playground about how to stop a traverse until a condition is met here: https://github.com/TristanCacqueray/haskell-playground/blob/master/mtl/TraverseMaybeT.hs

Perhaps we could a MaybeT here to avoid going through every workspace?

protos/monocle/http.proto
```diff
@@ -14,6 +14,17 @@ import "monocle/config.proto";
 import "monocle/search.proto";
 import "monocle/crawler.proto";
 import "monocle/user_group.proto";
+import "monocle/login.proto";
+
+service Login {
+  // Check a login username
+  rpc LoginValidation(monocle_login.LoginValidationRequest) returns (monocle_login.LoginValidationResponse) {
+    option (google.api.http) = {
+      post: "/api/2/login/username/validate"
```
TristanCacqueray, line 23: Why not using a shorter `/api/2/login` instead?

TristanCacqueray (approved):

TristanCacqueray (approved):

haskell/src/Monocle/Api/Server.hs
morucci: Yes sure that would be more efficient !

protos/monocle/http.proto
morucci: Well ok, I'll publish a followup.

PR #773, opened 2021-12-06
[ci] Mergify - update config due to deprecation

morucci:
See https://blog.mergify.com/strict-mode-deprecation/


morucci:
I guess this will need to be merged / w/o Mergify validation.

PR #778, opened 2021-12-07
[graphQL] client - honor the resetAt date

morucci:


morucci:
@TristanCacqueray Here is a draft proposal, I haven't properly tested it but I send it here for early comments about the implementation.

morucci:
Will rework it on top of #780

PR #781, opened 2021-12-08
[ci] mergify rename "Haskel GHC" test to "tests"

morucci:


morucci:
Attempt to fix https://github.com/change-metrics/monocle/pull/780 stick in mergify queue.

PR #782, opened 2021-12-08
[macroscope] graphQL - honor the resetAt date in the Stream

morucci:


morucci:
@TristanCacqueray let me know what do you think. I haven't tested it yet by the way.

haskell/src/Lentille/GraphQL.hs
```diff
@@ -129,4 +161,5 @@ streamFetch client mkArgs transformResponse = go Nothing
       unless (null decodingErrors) (stopLentille $ DecodeError decodingErrors)
 
       -- TODO: implement throttle
```
TristanCacqueray, line 163: I think you can remove the TODO

haskell/src/Monocle/Class.hs
```diff
@@ -101,7 +101,7 @@ instance MonadLog IO where
 -------------------------------------------------------------------------------
 -- A http system
 
-class (MonadRetry m, MonadLog m) => MonadGraphQL m where
+class (MonadRetry m, MonadLog m, MonadUnliftIO m) => MonadGraphQL m where
```
TristanCacqueray, line 104: Ideally we shouldn't enable IO, but we would need to define a class for `newMVar` and `modifyMVar`. But perhaps it is not so useful to have those classes and we could use a simpler `MonadUnlinftIO` constraint.

Thus I suggest we comment here that this is needed for Concurrent.MVar, and then in a follow-up we can clean-up this module.

TristanCacqueray (approved):

haskell/src/Lentille/GraphQL.hs
morucci: Python implementation got 1s delay between each API calls: https://github.com/change-metrics/monocle/blob/master/monocle/github/graphql.py#L86
I guess we will need more experimentation about crawling large repositories to see how this graphQL crawler implementation behaves with Github.

haskell/src/Monocle/Class.hs
morucci: Yes actually I started to do that in my first implementation. So I bring it back. I think that better now.

haskell/src/Monocle/Class.hs
```diff
@@ -24,6 +24,13 @@ instance MonadTime IO where
   mGetCurrentTime = Data.Time.Clock.getCurrentTime
   mThreadDelay = Control.Concurrent.threadDelay
 
+-------------------------------------------------------------------------------
+-- A concurrent system handled via Control.Concurrent.MVar (Unlifted from IO)
+
+class MonadUnliftIO m => MonadSync m where
```
TristanCacqueray, line 30: By having the constraint here makes IO available to any MonadSync user. I don't think it matter for this particular class, thus I would recommend to remove ths constraint.

And we should provide an instance for IO, so that we can keep the IO instance of MonadGraphQL below:

```haskell
class MonadSync IO where
  mNewMVar = newMVar
  mModifyMVar = modifyMVar
```

TristanCacqueray (approved):
That looks great!

PR #788, opened 2021-12-13
[macroscope] Add the Lentille.GitHub.Organization module

morucci:


haskell/src/Monocle/Api/Server.hs
```diff
@@ -287,7 +287,12 @@ crawlerAddDoc request = do
       pure $ CrawlerPB.AddDocResponse Nothing
     addProjects crawler organizationName projects = do
       monocleLogEvent $ AddingProject (getWorkerName crawler) organizationName (length projects)
-      I.initCrawlerEntities (Project <$> projectNames projects) crawler
+      let names = projectNames projects
+          -- TODO(fbo) do not crawl github issues by default for an organization.
```
TristanCacqueray, line 291: It seems like this is already the case, shouldn't the comment be: `TODO: figure out a strategy to crawl the issues` ?

TristanCacqueray (approved):

haskell/src/Monocle/Api/Server.hs
morucci: Right, I've updated the TODO comment

PR #789, opened 2021-12-13
Ensure (TLS enabled) mkManager is available for MonadGraphQL in IO

morucci:


haskell/src/Monocle/Client.hs
```diff
@@ -39,15 +39,15 @@ data MonocleClient = MonocleClient
   }
 
 -- | Create a HTTP manager
-mkManager :: MonadIO m => m Manager
+mkManager :: IO Manager
```
TristanCacqueray, line 42: I don't think that is necessary, if this was done for the MonadGraphQL IO instance, then remember that IO implements the MonadIO, so it should work similarly.

```suggestion
mkManager :: MonadIO m => m Manager
```

TristanCacqueray (approved):

PR #790, opened 2021-12-14
[macroscope] Migrate GitHub crawler to the new API

morucci:
This change migrates the Github crawler under the Macroscope engine (like other crawlers). Also as the legacy crawler was the the only remaining part requiring Python then this changes clean the Python bits from the Monocle repository.

- [x] Provide an iso-functional crawler compare to legacy one in Python
- [x] Validate that we gather the exact same data for some given repos (compare with the legacy crawler)
- [x] Validate that the crawler can fetch large repos w/o ratelimit/timeout issues
- [x] Ensure that after a crawler run (with new code) on a previous (with legacy code) does not duplicate docs
- [x] Ensure crawler metadata are created automatically
- [x] Remove legacy crawler support 
- [x] Attempt an upgrade in a sandbox with production data


haskell/src/Lentille/GraphQL.hs
```diff
@@ -110,17 +136,55 @@ data RateLimit = RateLimit {used :: Int, remaining :: Int, resetAt :: UTCTime}
...
   from RateLimit {..} = "remains:" <> show remaining <> ", reset at: " <> show resetAt
 
+type RetryCheck m a = (Either (FetchError a) a, [ReqLog]) -> m Bool
+
+-- | wrapper around fetchWithLog than can optionaly handle fetch retries
+-- based on the returned data inspection via a provided function (see RetryCheck).
+-- In case of retry the depth parameter of mkArgs is decreased (see adaptDepth)
+doRequest ::
+  forall a.
+  forall m.
```
TristanCacqueray, line 146: nit: a single forall can define multiple type variable, e.g. 
```suggestion
  forall a m.
```

haskell/src/Lentille/GraphQL.hs
```diff
@@ -110,17 +136,55 @@ data RateLimit = RateLimit {used :: Int, remaining :: Int, resetAt :: UTCTime}
...
+  (MonadGraphQLE m, Fetch a, FromJSON a) =>
+  GraphClient ->
+  (Maybe Int -> Maybe Text -> Args a) ->
+  Maybe (RetryCheck m a) ->
+  Maybe Int ->
+  Maybe PageInfo ->
```
TristanCacqueray, line 152: nit: when parameters are not likely to be pre-applied (e.g. curry), then they can be named using record field, e.g.:

```haskell
data RequestParams m a = RequestParams {
    fpMkArgs :: Maybe Int -> Maybe Text -> Args a
  , fpRetryCheck :: Maybe (RetryCheck m a)
  , ...
  }
```

Then the function would be: `doRequest :: GraphClient -> RequestParams m a -> m a`

And we can provide a `defaultRequestsParams = RequestParams Nothing Nothing Nothing`

haskell/src/Lentille/GraphQL.hs
```diff
@@ -110,17 +136,55 @@ data RateLimit = RateLimit {used :: Int, remaining :: Int, resetAt :: UTCTime}
...
+  m (Either (FetchError a) a, [ReqLog])
+doRequest client mkArgs retryCheckM depthM pageInfoM = gRetry retryCheck runFetch
+  where
+    gRetry = genericRetry Macroscope "Retrying request with smaller depth"
+    retryCheck _ = fromMaybe (const $ pure False) retryCheckM
+    runFetch :: (MonadGraphQLE m) => Int -> m (Either (FetchError a) a, [ReqLog])
+    runFetch retried = do
+      let adapedDepth = adaptDepth
+      if adapedDepth == Just 0
```
TristanCacqueray, line 161:
```suggestion
      if adaptDepth == Just 0
```

haskell/src/Lentille/GraphQL.hs
```diff
@@ -110,17 +136,55 @@ data RateLimit = RateLimit {used :: Int, remaining :: Int, resetAt :: UTCTime}
...
+        then error "Unable to reduce depth more"
+        else
+          fetchWithLog
+            (doGraphRequest client)
+            (mkArgs depthM $ (Just . fromMaybe (error "Missing endCursor from page info") . endCursor) =<< pageInfoM)
+      where
+        adaptDepth :: Maybe Int
+        adaptDepth =
+          let decValue = truncate @Float . (* (fromIntegral retried * 0.3)) . fromIntegral <$> depthM
+           in (-) <$> depthM <*> decValue
```
TristanCacqueray, line 171: might be good to factor the non monadic part as a top-level function with some doctest such as:

```suggestion
        adaptDepth :: Maybe Int
        adaptDepth = decreaseValue retried <$> depthM
        
-- | Slowly decrease a value to workaround api timeout when a graph depth is too deep.
 -- >>> decreaseValue 1 42
 -- 29
 -- >>> decreaseValue 1 29
 -- 20
decreaseValue :: Int -> Int -> Int
decreaseValue retried depth =
          let decValue = truncate @Float . (* (fromIntegral retried * 0.3)) . fromIntegral $ depth
           in depth - decValue
```

TristanCacqueray: Shouldn't the decreaseValue also clamp at 0?, e.g. `max 0 $ depth - decValue`

haskell/src/Lentille/GraphQL.hs
```diff
@@ -110,17 +136,55 @@ data RateLimit = RateLimit {used :: Int, remaining :: Int, resetAt :: UTCTime}
...
+  m (Either (FetchError a) a, [ReqLog])
+doRequest client mkArgs retryCheckM depthM pageInfoM = gRetry retryCheck runFetch
+  where
+    gRetry = genericRetry Macroscope "Retrying request with smaller depth"
+    retryCheck _ = fromMaybe (const $ pure False) retryCheckM
+    runFetch :: (MonadGraphQLE m) => Int -> m (Either (FetchError a) a, [ReqLog])
+    runFetch retried = do
+      let adapedDepth = adaptDepth
+      if adapedDepth == Just 0
+        then error "Unable to reduce depth more"
```
TristanCacqueray, line 162: since error is a bottom, you can also replace the if/then/else with:

```suggestion
        when (adapedDepth == just 0) $ error "Unable to reduce depth more"
```

haskell/src/Lentille.hs
```diff
@@ -94,6 +106,7 @@ instance MonadLog LentilleM where
 
 instance MonadRetry LentilleM where
   retry = retry'
+  genericRetry = genericRetry'
```
TristanCacqueray, line 109: would have been nice if retry was implemented in term of genericRetry so that the MonadRetry only need a single method. But that's ok like that too.

TristanCacqueray (commented):
I haven't tried the code, but the implementation looks good to me. I only found a few nits that can be addressed in follow-up. Well done!

TristanCacqueray (approved):

haskell/src/Lentille.hs
morucci: Yes true, I'll let a TODO note for that. An awesome task for a rainy Sunday .... ;)

haskell/src/Lentille/GraphQL.hs
morucci: I've set a max to 1 so I'm removing the exception raise here. It's unlikely  to happen that a single PR cannot be fetched.

haskell/src/Lentille/GraphQL.hs
morucci: Thanks you for the tips. As of now doRequest is only used by streamFetch so it does not make so much sense to do it there. However streamFetch might be a better candidate.

morucci: This has been implemented in a81b8b01dfbea551926f5734819dbd73cde340a6

haskell/src/Lentille.hs
TristanCacqueray: Well the retry class is really about an abstraction for the `retry` library, e.g. to be able to catch and re-throw exception. Thus I think we should, on a rainy Sunday, try to implement the http retry in term of genericRetry so that the class only has a single retry function. But that's more of a nit, it's ok like what you proposed.

haskell/src/Lentille/GitHub/PullRequests.hs
```diff
@@ -0,0 +1,553 @@
...
+                  (ChangeEventTypeChangeCommented ChangeCommentedEvent)
+                  ("ChangeCommmentedEvent-" <> changeId change <> "-" <> from eId)
+                  change
+              )
+                { changeEventAuthor = Just $ getIdent actor,
+                  changeEventCreatedAt = Just $ from createdAt
+                }
+        toEvent _ = Nothing
+    transPR :: SearchNodesSearchResultItem -> Maybe (Change, [ChangeEvent])
+    transPR (SearchNodesSearchResultItem _) = Nothing
```
TristanCacqueray, line 485: Shouldn't this case be handled?

haskell/src/Lentille/GitHub/PullRequests.hs
```diff
@@ -0,0 +1,553 @@
...
+            Just
+              ( baseEvent
+                  (ChangeEventTypeChangeCommented ChangeCommentedEvent)
+                  ("ChangeCommmentedEvent-" <> changeId change <> "-" <> from eId)
+                  change
+              )
+                { changeEventAuthor = Just $ getIdent actor,
+                  changeEventCreatedAt = Just $ from createdAt
+                }
+        toEvent _ = Nothing
```
TristanCacqueray, line 483: And what about this case?

morucci: For this one removing the line raises a warning about an incomplete pattern match. However the first and unique match is the one we need. So the default match seems to be needed to satisfied the type checker.

TristanCacqueray:
Perhaps we could copy the prdata fragment to another test query, and we should be able to re-use the transform response by using [unsafeCoerce](https://hackage.haskell.org/package/base-4.16.0.0/docs/Unsafe-Coerce.html#v:unsafeCoerce) so that we can run the code for a single PR using the search query.

TristanCacqueray:
Or perhaps even in the same module we could have two `query`, e.g. `GetProjectPullRequests` and `GetProjectPullRequest`, where the former use `repository` and the later use `search`?

morucci:
> Perhaps we could copy the prdata fragment to another test query, and we should be able to re-use the transform response by using [unsafeCoerce](https://hackage.haskell.org/package/base-4.16.0.0/docs/Unsafe-Coerce.html#v:unsafeCoerce) so that we can run the code for a single PR using the search query.

Yes I tried that using coerce (Data.Coerce) but that was a failure due to type mismatch, so yes maybe retrying with unsafe coerce might worth the try.

morucci:
> Or perhaps even in the same module we could have two `query`, e.g. `GetProjectPullRequests` and `GetProjectPullRequest`, where the former use `repository` and the later use `search`?

Both can use repository but sub functions must be for the first one PullRequest and PullRequests for the second. Ok but it will require a almost full duplicate of code.

TristanCacqueray:
About migrating crawler metadata, here is what I suggest:

- Let's create a `audit` index, and if read the `version` document id
- If the document does not exists, lets assume we are at version 0

Then we can handle automatic migration, for example, from 0 to 1 it would be:

- For each github project, get the last updated date and update the crawler metadata with the date
- Then update the audit/version document to version 1

CHANGELOG.md
```diff
@@ -4,15 +4,26 @@ All notable changes to this project will be documented in this file.
 
 ## [master]
 
+This release mainly removes a technical dept where the GitHub Pull-Requests crawler
+was managed differently than other crawlers. With this release, the crawler is managed
+by the same process (or container) than other crawlers and respects the same Macroscope
```
TristanCacqueray, line 9: I think we should keep the `Macroscope` and `Lentille` names for the developper documentation, that should not be visible to the end user.

```suggestion
by the same process (or container) than other crawlers and respects the same
```

TristanCacqueray (approved):

CHANGELOG.md
morucci: done

CHANGELOG.md
```diff
@@ -4,15 +4,26 @@ All notable changes to this project will be documented in this file.
...
+Monocle configuration or database schema updates.
+
 ### Added
 
 - [web] A minimal login capability to enable the use 'self' keyword in queries.
 - [cli] --version argument to display the current version
 
 ### Changed
+
+- [crawler] GitHub Pull Request crawler has been migrated under Macroscope.
```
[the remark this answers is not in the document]
morucci: done

morucci:
Ok that looks good to me. Just a couple of comments, also see that CI is reporting a possible ETA reduction.

morucci:
I'll followup and fill upgradeConfigV1  ! Thanks

haskell/src/Monocle/Backend/Index.hs
```diff
@@ -278,11 +279,50 @@ configDoc = BH.DocId "config"
 -- | Upgrade to config v1 (migrate legacy GH crawler to the new API)
 upgradeConfigV1 :: QueryM ()
```
morucci, line 280: I'll add test case for it in a follow-up commit.

haskell/src/Monocle/Backend/Index.hs
```diff
@@ -456,9 +486,8 @@ checkDocExists docId = do
   index <- getIndexName
   BH.documentExists index docId
 
-getDocumentById :: (FromJSON a) => BH.DocId -> QueryM (Maybe a)
-getDocumentById docId = do
-  index <- getIndexName
+getDocumentById' :: (BH.MonadBH m, FromJSON a, MonadThrow m) => BH.IndexName -> BH.DocId -> m (Maybe a)
```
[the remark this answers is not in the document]
TristanCacqueray: right, i did the split for the first version which out the query context (ensureConfig was using the MonadBH). We can revert back that modification indeed.

haskell/monocle.cabal
```diff
@@ -128,6 +128,7 @@ library
                      , http-types                 >=0.12
                      , json-syntax                >= 0.2
                      , lens
+                     , lens-aeson
```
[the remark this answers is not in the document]
TristanCacqueray: I wanted to show how we can handle raw Value efficiently with these lenses, and I think it's good to know how that works. But I agree, it would be better to create a ConfigDocument data type with a FromJSON instance.

haskell/src/Monocle/Backend/Index.hs
TristanCacqueray: Thanks. for the test case I think we should force the config version to 0 by directly calling the indexDocument function.

TristanCacqueray (approved):

PR #794, opened 2021-12-21
fix(actions): use upstream action

georgettica:
the upstream repo that was used incorporated required changes to change
the version of minikube used.

this allows running using a newer version then previously expected


morucci:
Hi @georgettica , could you set a commit message to reflect the reason of the change ?

georgettica:
@morucci please review changes to commit and PR :)

morucci (approved):

morucci:
Thanks !

PR #795, opened 2021-12-23
feat(k8s): use stable versions

georgettica:
now we don't need to worry about touching the action versions anymore (until it breaks something on our end

### DIFF
--- a/.github/workflows/k8s.yaml
+++ b/.github/workflows/k8s.yaml
@@ -27,9 +27,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
       - name: Start minikube
-        uses: medyagh/setup-minikube@v0.0.3
+        uses: medyagh/setup-minikube@v0.0.7
         with:
-          minikube-version: 1.24.0
+          minikube-version: latest
+          kubernetes-version: stable
       - name: run k8s deployment on the cluster
         run: |
           kubectl create namespace "${namespace}" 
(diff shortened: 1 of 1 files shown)


georgettica:
seems they released a version that returns a 404 😱

.github/workflows/k8s.yaml
```diff
@@ -27,9 +27,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
       - name: Start minikube
-        uses: medyagh/setup-minikube@v0.0.3
+        uses: medyagh/setup-minikube@0ff5b450588f2aee88fd0daba616d04a5582e490
```
medyagh, line 30: this needs to be
medyagh/setup-minikube@v0.0.7
or 
medyagh/setup-minikube@latest

here is example:
https://github.com/medyagh/test-minikube-example/pull/7

https://github.com/medyagh/test-minikube-example/runs/4621762363?check_suite_focus=true

medyagh (changes requested):

.github/workflows/k8s.yaml
georgettica: As expected, doesn't work 😢 
https://github.com/change-metrics/monocle/runs/4623008565?check_suite_focus=true

georgettica: Also removing the k8s-version didn't work :(

georgettica: Seems moving from stable to latest works

georgettica:
I'll squash them Sunday.
No need to merge like this

morucci:
Hi @georgettica thanks for the change. Would you like to squash them ?
Should we wait to be able to use stable instead of latest ?

georgettica:
Actually, using stable is good enough.. the docs are not as clear as they should and this is the format taking the most recent version of k8s and minikube for testing

I'll squash tomorrow

georgettica:
@morucci squashed now, let's hope it passes in a first try

morucci (approved):

PR #805, opened 2022-01-26
[logging] complete removal of Say package

morucci:


haskell/src/Monocle/Prelude.hs
```diff
@@ -426,8 +422,13 @@ inText sub txt = case indices sub txt of
   [] -> False
   _ -> True
 
+monocleLogIO :: Text -> IO ()
+monocleLogIO log = liftIO $ withLogger callLogger
+  where
+    callLogger logger = doLog logger $ encodeUtf8 @Text @ByteString . from $ log
+
```
TristanCacqueray, line 429: We shouldn't need the unlifted version, and we might want to keep the bytestring version. How about:

```suggestion
-- | Print a message to the stderr with a timestamp
logMessage :: MonadIO m => ByteString -> m ()
logMessage msg = liftIO $ withLogger (flip doLog msg)

-- | Print a text message to the stderr
logText :: MonadIO m => Text -> m ()
logText = logMessage . from
```

TristanCacqueray (approved):

haskell/src/Monocle/Prelude.hs
morucci: Thanks you ! done

PR #808, opened 2022-01-27
[DOC] Some improvements on nix commands/README.md/CONTRIBUTING.md

morucci:
- Renamed monocle-api2-start to monocle-repl
- Cleaning haskell/README.md as it was a duplicate of CONTRIBUTING.md
- Some slight re-org and intro improvement
- Reflecting nix changes into the CONTRIBUTING.md file
- Replace instruction regarding start-web.sh that has been removed from
  contrib/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,13 +1,16 @@
 # Contributing
 
 This document provides some instructions to get started with Monocle development. It covers topics
-such as running tests, running services w/o docker-compose or running the codegen.
+such as running tests, running services or running the codegen.
 
 ## Understanding the design choices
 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -163,13 +157,27 @@ firefox http://localhost:13000
...
+```
+
+#### Run all components in emac buffers
+
+Note that the commands below can be started in emacs buffers using:
+
+```ShellSession
+nix-shell --command launch-monocle-with-emacs
+```
+
```
TristanCacqueray, line 180: well, i think we can skip that command :-) 

```suggestion
```

morucci: Done

PR #812, opened 2022-01-28
Avoid infinite loop in Macroscope in case of last buggy entity

morucci:


haskell/src/Monocle/Prelude.hs
```diff
@@ -410,6 +410,9 @@ instance Num Count where
 instance From Int Int32 where
   from = fromInteger . toInteger
 
+instance From Word32 Int where
+  from = fromInteger . toInteger
```
TristanCacqueray, line 414: That looks like an omission from the witch library, would you like to propose this instance upstream?

For Int to Int32, I think we should be using tryfrom to avoid overflow.

TristanCacqueray (approved):
recheck

haskell/src/Monocle/Prelude.hs
morucci: Ok I'll propose a change upstream to the Witch project.
For Int to Int32 a'll tackle that in another change.

PR #813, opened 2022-01-28
Refactor the MonadRetry

TristanCacqueray:
This PR unifies the retry facilities to simplify the code.

### DIFF
--- a/haskell/src/Lentille/GraphQL.hs
+++ b/haskell/src/Lentille/GraphQL.hs
@@ -236,9 +236,11 @@ streamFetch client@GraphClient {..} lc mkArgs StreamFetchOptParams {..} transfor
           <> " - "
           <> maybe "no ratelimit" show rateLimitM
 
+    retryDelay = 1_100_000
+
     go pageInfoM totalFetched = do
-      --- Start be waiting one second if we request a new page
-      when (isJust pageInfoM) $ lift $ mThreadDelay 1_000_000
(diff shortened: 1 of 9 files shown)


haskell/src/Lentille/GraphQL.hs
```diff
@@ -227,7 +238,7 @@ streamFetch client@GraphClient {..} lc mkArgs StreamFetchOptParams {..} transfor
 
     go pageInfoM totalFetched = do
       --- Start be waiting one second if we request a new page
-      when (isJust pageInfoM) $ lift $ mThreadDelay remoteCallDelay
+      when (isJust pageInfoM) $ lift $ mThreadDelay 1_000_000
```
morucci, line 241: let's keep the previous delay of 1.1 second (I feel it's important ...)

haskell/src/Lentille/GraphQL.hs
```diff
@@ -236,7 +247,7 @@ streamFetch client@GraphClient {..} lc mkArgs StreamFetchOptParams {..} transfor
             const $ do
               rl <- getRateLimit client
               -- Wait one second to delay the next call
-              mThreadDelay remoteCallDelay
+              mThreadDelay 1_000_000
```
morucci, line 250: Ditto

haskell/src/Lentille/GraphQL.hs
morucci: I've pushed a new commit

PR #818, opened 2022-01-31
[crawler] Add Module Lentille.GitHub.User

morucci:
Usage will be to be able to enhance Identities into Monocle with extra
info fetched from a GitHub provider.

### DIFF
--- a/haskell/src/Lentille/GitHub/User.hs
+++ b/haskell/src/Lentille/GitHub/User.hs
@@ -1,0 +1,54 @@
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-overlapping-patterns #-}
+
+module Lentille.GitHub.User where
+
+import Data.Morpheus.Client
+import Lentille
+import Lentille.GitHub.RateLimit (retryCheck)
+import Lentille.GraphQL
+import Monocle.Prelude
+
+newtype DateTime = DateTime Text deriving (Show, Eq, EncodeScalar, DecodeScalar)
+
+data IdentInfo = IdentInfo
+  { iiName :: Maybe Text,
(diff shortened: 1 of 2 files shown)


haskell/src/Lentille/GitHub/User.hs
```diff
@@ -0,0 +1,54 @@
...
+import Data.Morpheus.Client
+import Lentille
+import Lentille.GitHub.RateLimit (retryCheck)
+import Lentille.GraphQL
+import Monocle.Prelude
+
+newtype DateTime = DateTime Text deriving (Show, Eq, EncodeScalar, DecodeScalar)
+
+data IdentInfo = IdentInfo
+  { iiName :: Maybe Text,
```
TristanCacqueray, line 17: Shouldn't this be:

```suggestion
  { iiName :: Text,
```

haskell/src/Lentille/GitHub/User.hs
```diff
@@ -0,0 +1,54 @@
...
+        resetAt
+      }
+      user(login: $login) {
+        name
+        company
+      }
+    }
+  |]
+
+transformResponse :: GetUser -> (Maybe RateLimit, IdentInfo)
```
TristanCacqueray, line 37:
```suggestion
transformResponse :: GetUser -> (RateLimit, Maybe IdentInfo)
```

haskell/src/Lentille/GitHub/User.hs
morucci: The schema binds to a Maybe Text

haskell/src/Lentille/GitHub/User.hs
morucci: Yes, I've removed the Maybe RateLimit. However we should get a IdentInfo or an exception will be raised, so for me that's an IdentInfo but not a Maybe IdentInfo.

morucci:
@TristanCacqueray could I get your approval ?

TristanCacqueray (approved):

PR #823, opened 2022-02-13
[crawler] Use openssl for graphql client

### LINKED ISSUE #822: SSL Certificate: Invalid header type 60
I am seeing a weird message in my crawler logs

```
crawler_1  | 2022-02-13 09:14:31 Unknown: Network error: 4/7 github.mycomp.com:443/ failed: InternalException 
(HandshakeFailed (Error_Packet_Parsing "Failed reading: invalid header type: 60\nFrom:\theader\n\n"))
```

I am using GitHub enterprise, and I am interesting in a single repo which is at URL `https://github.mycomp.com/tech-eng/myrepo`
The config I am using is 

```yaml
workspaces:
  - name: myworkspace
    crawlers:
      - name: mycompany-crawler
        update_since: '2019-01-01'
        provider:
          github_organization: tech-eng
          github_url: https://github.mycomp.com
          github_token: GITHUB_TOKEN
          github_repositories:
            - myepo
```

And the token in the secrets is my valid token. I am wondering if it might be a certificate issue as the ssl certificate is not public, so the container might not be able to validate it. If so, is there any option I could use to accept invalid certificate? or to upload another cert authority?

TristanCacqueray:
Fixes #822

### DIFF
--- a/haskell/monocle.cabal
+++ b/haskell/monocle.cabal
@@ -97,6 +97,7 @@ common codegen
 library
   import:              common-options, codegen
   build-depends:       base                       < 5
+                     , HsOpenSSL                  >= 0.11
                      , MonadRandom
                      , aeson
                      , aeson-casing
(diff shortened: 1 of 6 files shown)


morucci:
Ok so I assume we'll get less weird behaviors by using the OpenSSL binding instead of the http-client-tls based on cryptonite. I did a quick test by crawling on github.com and that went fine. So let's move on with this change.
Thanks @TristanCacqueray

PR #826, opened 2022-02-14
[crawler] Enable settings TLS_CIPHERS openssl option

### LINKED ISSUE #822: SSL Certificate: Invalid header type 60

TristanCacqueray:
This change enables setting https://www.openssl.org/docs/man1.1.1/man1/ciphers.html

Fixes #822

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [master]
 ### Added
+
+- [crawler] A new env var "TLS_CIPHERS" to override the default OpenSSL cyphers list.
+
 ### Changed
 ### Removed
 ### Fixed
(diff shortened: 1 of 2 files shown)


morucci (approved):
That looks good to me. I'll add a changelog entry for this before merging.

Issue #764, opened 2021-12-01
Author activity page


TristanCacqueray:
A new activity page design proposal:

```
(when personalized
   |----------------------------|
   | Gravatar:
   | Member of groups:
   |----------------------------|
)

|--Stats----------------------------------------|   |--Projects-------------------------------------|
| Total Commits: XX                             |   | zuu/zuul: XX        (previous: +/- YY)        |
| Total Reviews: XX   (previous period: +/- YY) |   | ....                                          |
| Total Issues:  XX                             |   |-----------------------------------------------|
| Total Open:    XX                             |
| Total Merges:  YY                             |
|-----------------------------------------------|

|---------------------------------------------------------------------------------------------------|
| Reviews: [] [] [] [] [] [] [] [] [] [] []
| Commits: [] [] [] [] [] [] [] [] [] [] []
|---------------------------------------------------------------------------------------------------|

|--Event list------[ ] filter event type/comment/merged/...-----------------------------------------|
| [pr#1]
| [pr#2]
|-----------------------------------------------|
```

morucci:
I'll try to propose a new page with items and info similar to the mock-up. I think this page should be generic enough (if possible) to fit an individual or a group.

PR #827, opened 2022-02-15
Add Author/Group scoped view

morucci:
Related-to #764


morucci:
This change is ready for testing. Once merged, as next steps we could think of:

1. extracting group members metrics as suggested here: #830
2. adding a search authors page

TristanCacqueray (approved):
That looks solid. I'd like to give it a try, but I don't have a dev setup at the moment. Feel free to merge it.

PR #831, opened 2022-02-22
Updates to the readme

LunaticMuch:
This PR addresses the following:

1. The github url does not clarify that the URL must include the path to the graphql endpoint. This is made clear on `Gitlab` section only; so I am proposing some consistency
2. The troubleshooting section includes a note about possible issues with SSL certificate which are resolved, but not documented. Because they are fortunately not common, I propose to have a note on troubleshooting section only. 
#822

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ A GitHub provider settings
     # Optional settings
     github_repositories:
       - pipeline
-    github_url: https://github.com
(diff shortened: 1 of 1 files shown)


TristanCacqueray (approved):
That looks great, thank you for taking the time to update the documentation!

PR #844, opened 2022-03-16
Add an API endpoint to search through event authors

morucci:


haskell/src/Monocle/Backend/Index.hs
```diff
@@ -206,6 +220,9 @@ createIndex indexName mapping = do
   where
     indexSettings = BH.IndexSettings (BH.ShardCount 1) (BH.ReplicaCount 0)
 
+schemaVersion :: ConfigVersion
```
TristanCacqueray, line 223: nit: perhaps this should be named `configVersion`, or `latestConfigVersion` ?

haskell/src/Monocle/Backend/Index.hs
```diff
@@ -294,10 +321,14 @@ ensureConfigIndex = do
   createIndex configIndex ConfigIndexMapping
   (currentVersion, currentConfig) <- getConfigVersion
 
-  when (currentVersion == ConfigVersion 0) $ traverseWorkspace upgradeConfigV1 conf
+  when (currentVersion == ConfigVersion 0) $ do
+    traverseWorkspace upgradeConfigV1 conf
+    traverseWorkspace upgradeConfigV2 conf
```
TristanCacqueray, line 326: If this gets too complicated to maintain, I think we could define upgrade procedures like this:

```haskell
upgradeConfig :: [(ConfigVersion, TenantM ())]
upgradeConfig = [
    (ConfigVersion 1, do
     -- ensure crawler updated_at
     undefined
  ), (ConfigVersion 2, do
     --  add cached author
     undefined
  )]
```

So that we can automatically run the migration that are above the current version. With something like `traverse (\(version, procedure) -> when (version < currentVersion) procedure) upgradeConfig`

TristanCacqueray (approved):

haskell/src/Monocle/Backend/Index.hs
morucci: Thanks see next commit.

haskell/src/Monocle/Backend/Test.hs
```diff
@@ -27,14 +27,16 @@ import qualified Monocle.Search.Query as Q
 import Relude.Unsafe ((!!))
 import qualified Streaming.Prelude as Streaming
 
-fakeDate :: UTCTime
+fakeDate, fakeDateAlt :: UTCTime
```
TristanCacqueray, line 30: That's much better!

haskell/src/Monocle/Backend/Test.hs
```diff
@@ -951,11 +953,15 @@ testAuthorCache = withTenant doTest
...
             { caType = ECachedAuthor,
-              caCachedAuthorMuid = "bob"
+              caCachedAuthorMuid = "John Doe/12"
             }
         ]
         resp
 
+      -- Validate that we can search in the author cache
+      authors <- I.searchAuthorCache "doe"
+      assertEqual' "Check search author" ["John Doe/12"] authors
```
TristanCacqueray, line 963: well done!

TristanCacqueray (approved):

PR #851, opened 2022-03-25
[wabapp] make the search authors page to display "No results"

morucci:
After an un-successful query we need to display "No results".


TristanCacqueray (approved):

morucci:
Thanks for the review !

PR #852, opened 2022-03-25
[clean + haddock] Monocle.Api.Config

morucci:


haskell/src/Monocle/Prelude.hs
```diff
@@ -435,6 +436,9 @@ getExn (Left err) = error (toText err)
 stripSpaces :: Text -> Text
 stripSpaces = T.replace " " ""
 
+removeTrailingSlash :: Text -> Text
+removeTrailingSlash = T.dropWhileEnd (== '/')
```
TristanCacqueray, line 440: I think there is a rule that says that if the name of function is as long as the expression, then it is not worth defining it... Perhaps this should be part of a `Text -> ApiURL` function?

haskell/src/Monocle/Test/Spec.hs
```diff
@@ -569,8 +569,37 @@ monocleConfig :: TestTree
...
+      testGetTenantGroups,
+      testGetIdentByAlias
+    ]
   where
+    createIdent :: Text -> [Text] -> [Text] -> Ident
+    createIdent name aliases' groups' =
+      let ident = name
+          aliases = aliases'
+          groups = Just groups'
+       in Ident {..}
```
TristanCacqueray, line 582: Why not?

```suggestion
    createIdent ident aliases groups' =
      let groups = Just groups'
       in Ident {..}
```

TristanCacqueray (approved):

morucci:
Thanks for the review !

PR #858, opened 2022-03-31
[nix] Update haskell dependencies

TristanCacqueray:


TristanCacqueray:
This seems to build fine with cabal locally too, so I think we should skip the PR validation CI, and rebuild the builder image after this is merged.

morucci (approved):

PR #862, opened 2022-04-05
chore(web): bump chokidar

georgettica:
solved a security vuln
chokidar 2 does not recieve security updates anymore

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -12,6 +12,7 @@
     "bs-ocaml-protoc-json": "^0.1.3",
     "bs-parse": "git+https://github.com/aitoroses/bs-parse.git",
     "chart.js": "^2.9.4",
+    "chokidar": "^3.5.3",
     "concurrently": "^6.0.2",
     "interweave": "^12.7.1",
     "interweave-autolink": "^4.4.1",
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -291,6 +291,15 @@
         "color-convert": "^2.0.1"
       }
     },
+    "anymatch": {
+      "version": "3.1.2",
+      "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-3.1.2.tgz",
+      "integrity": "sha512-P43ePfOAIupkguHUycrc4qJ9kz8ZiuOUijaETwX7THt0Y/GNK7v0aa8rY816xWjZ7rJdA5XdMcpVFTKMq+RvWg==",
+      "requires": {
+        "normalize-path": "^3.0.0",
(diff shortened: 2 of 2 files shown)


georgettica:
local branch passed GHA, should pass here as well :)

georgettica:
and it did :)
hope this get's merged and it will solve my local security error

morucci:
Thanks for the PR.

I've tried it, then
```
npm install
npm audit
```
but still I get the issue:

```
[fedora@localhost web]$ npm audit
# npm audit report

glob-parent  <5.1.2
Severity: high
Regular expression denial of service in glob-parent - https://github.com/advisories/GHSA-ww39-953v-wcq6
fix available via `npm audit fix --force`
Will install live-server@1.1.0, which is a breaking change
node_modules/live-server/node_modules/glob-parent
  chokidar  1.0.0-rc1 - 2.1.8
  Depends on vulnerable versions of glob-parent
  node_modules/live-server/node_modules/chokidar
    live-server  >=1.2.0
    Depends on vulnerable versions of chokidar
    node_modules/live-server

3 high severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force
[fedora@localhost web]$ git log --oneline | head -1
f778d608 chore(web): bump chokidar
```

Furthermore It seems the issue is related to the live-server module that we only use for development. Should it be added to *devDependencies* section.

morucci:
It seems that the GH audit is reporting the same as npm audit (see the above output)
live-server@1.2.1 requires glob-parent@^3.1.0 via chokidar@2.1.8
Then I don't think that will clear the GH audit.

IMO the proper fix is to move out from the live-server package to something similar.

georgettica:
From the code I thought It changed to a newer version.. did that not happen?

georgettica:
Oh, it's because of glob-parent. Thanks!

web/package.json
```diff
@@ -12,6 +12,7 @@
     "bs-ocaml-protoc-json": "^0.1.3",
     "bs-parse": "git+https://github.com/aitoroses/bs-parse.git",
     "chart.js": "^2.9.4",
+    "chokidar": "^3.5.3",
```
TristanCacqueray, line 15: That's not correct because this is not a direct dependencies of monocle. To get a new version of that library, you need to find what is requiring it, and update the minimum version of one of the existing dependency.

georgettica: I see that now. I will update the PR with changing the base package

morucci:
I'll close this PR. Please re-open if needed.

PR #863, opened 2022-04-05
feat(migrator): add new component

georgettica:
this migrator will allow migration between versions of monocle without the need of downtime.
currently it's empty, but the code can be changed so anyone can run his migration steps on the repo

### DIFF
--- a/Dockerfile-migrator
+++ b/Dockerfile-migrator
@@ -1,0 +1,31 @@
+# Monocle.
+# Copyright (C) 2021 Monocle authors
+#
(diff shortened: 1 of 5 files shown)


georgettica:
this will be a draft as I am not sure how you would like to implement the cronjob. also this is currently untested so maybe some resources will need tweaking

georgettica:
I guess even if the image is missing this passes as it doesn't create a job from the cronjob (which is great)

need to add GHA tests for this (to create the job and see it works as expected)

TristanCacqueray:
Please re-open if needed.

PR #867, opened 2022-04-13
nix: update the proto3-suite reference

TristanCacqueray:
The fix was merged and the reference needs to be updated.


TristanCacqueray:
Fasttracking the fix to unbreak the CI.

PR #871, opened 2022-04-18
Add github watching query

TristanCacqueray:


haskell/src/Lentille/GitHub/Watching.hs
```diff
@@ -0,0 +1,93 @@
...
+
+module Lentille.GitHub.Watching where
+
+import Data.Morpheus.Client qualified
+import Lentille qualified
+import Lentille.GitHub.RateLimit qualified
+import Lentille.GraphQL qualified
+import Monocle.Logging qualified
+import Monocle.Prelude
+import Monocle.Protob.Crawler qualified
```
morucci, line 16: Fully qualified <3

morucci (approved):

morucci:
Perhaps add a changelog entry to tell that the CLI get a new command.

haskell/src/Lentille/GitHub/Watching.hs
TristanCacqueray: I think that's the way to go :)

TristanCacqueray (commented):
I meant to add a new crawler to add all the projects watched by a given user. But perhaps that could be done in a followup PR?

PR #878, opened 2022-05-23
Add base support in Monocle API to create/validate Magic JWT

morucci:
This change introduces two new endpoints:
- **auth/get**: By passing a valid admin_token a client can request a Magic JWT 
- **auth/whoami**: A client can validate a Magic JWT

The API must be started with the ADMIN_TOKEN to enable the auth/get endpoint.

All endpoints support the JWT authentication but none of them (except auth/whoami) check if the JWT is valid.

**This change is the prelude to the support of an external OIDC provider.**


haskell/src/Monocle/Api/Jwt.hs
```diff
@@ -0,0 +1,50 @@
...
+    (?~),
+  )
+import Relude
+
+mkMagicClaims :: MonadIO m => StringOrURI -> m ClaimsSet
+mkMagicClaims subject = do
+  t <- getCurrentTime
+  pure $
+    emptyClaimsSet
+      & claimIss ?~ "MagicJWT"
```
TristanCacqueray, line 34: This seems to be equivalent if you want to avoid adding a new lens operator.

```suggestion
      & claimIss `set` Just "MagicJWT"
```

morucci:
Need to check what's the issue with: provideOptions monocleAPI

TristanCacqueray:
> Need to check what's the issue with: "provideOptions monocleAPI". However, I'm still experimenting at that step :)

I think we need to find (or implement) the [GenerateList](https://hackage.haskell.org/package/servant-foreign-0.15.4/docs/Servant-Foreign.html#t:GenerateList) instance for Auth.JWT ... not sure how though :)

haskell/monocle.cabal
```diff
@@ -148,8 +149,9 @@ library
                      , retry                      >= 0.8
                      , safe-exceptions            >= 0.1
                      , servant                    >= 0.18.2
-                     , servant-options            >= 0.1
+                    --  , servant-options            >= 0.1
```
TristanCacqueray, line 152: Let's remove the servant-options requirement

haskell/src/Monocle/Client.hs
```diff
@@ -67,11 +69,13 @@ withClient ::
...
   -- | An optional manager
   Maybe Manager ->
+  -- | An option Auth token
+  Maybe Text ->
   -- | The callback
   (MonocleClient -> m a) ->
   -- | withClient performs the IO
   m a
-withClient url managerM callBack =
+withClient url managerM tokenM callBack =
```
TristanCacqueray, line 78: Should we read the token from the environment instead? Otherwise `withClient` user will all have to do that instead.

haskell/src/Monocle/Client.hs
```diff
@@ -89,7 +93,11 @@ monocleReq path MonocleClient {..} body =
     initRequest <- parseUrlThrow (T.unpack $ baseUrl <> path)
     let request =
           initRequest
-            { requestHeaders = [("Accept", "*/*"), ("Content-Type", "application/json")],
+            { requestHeaders =
+                [ ("Accept", "*/*"),
+                  ("Content-Type", "application/json"),
+                  authorizationH
+                ],
```
TristanCacqueray, line 100: not sure how that works, perhaps it would be easier to mappend a list, e.g.:

```suggestion
                  ("Content-Type", "application/json")
                ] <> Data.Maybe.maybeToList authorizationM,
```

haskell/src/Monocle/Client.hs
```diff
@@ -99,3 +107,4 @@ monocleReq path MonocleClient {..} body =
     decodeResponse body' = case JSONPB.eitherDecode body' of
       Left err -> error $ "Decoding of " <> show body <> " failed with: " <> show err
       Right a -> a
+    authorizationH = maybe mempty (\token -> ("Authorization", "Bearer " <> from token)) tokenM
```
TristanCacqueray, line 110:
```suggestion
    authorizationH = fmap (\token -> ("Authorization", "Bearer " <> from token)) tokenM
```

haskell/src/Monocle/Main.hs
```diff
@@ -1,7 +1,10 @@
 -- | The Monocle entry point.
 module Monocle.Main (run, app) where
 
+-- import Network.Wai.Middleware.Servant.Options (provideOptions)
+
```
TristanCacqueray, line 5:
```suggestion
```

haskell/src/Monocle/Main.hs
```diff
@@ -75,7 +87,9 @@ run' port url configFile glLogger = do
       Warp.runSettings
         settings
         . cors (const $ Just policy)
-        . provideOptions monocleAPI
+        -- https://github.com/haskell-servant/servant/issues/672
+        -- TODO: commented out for now due to missing instances
+        -- . provideOptions monocleAPI
```
TristanCacqueray, line 92:
```suggestion
```

If that doesn't work, I think we could add `corsIgnoreFailures = True` in the policy below. Though in my testing, the web development ui worked without provideOptions.

haskell/src/Monocle/Prelude.hs
```diff
@@ -126,6 +126,7 @@ module Monocle.Prelude
     preview,
     at,
     set,
+    (?~),
```
TristanCacqueray, line 129: It doesn't seem used anymore, and I think we should avoid adding these lens operator.

TristanCacqueray (approved):
:+1:

haskell/src/Monocle/Client.hs
morucci: maybeToList is useful ! right

haskell/src/Monocle/Main.hs
morucci: Oh yeah, I've retried today an yes simply removing the provideOptions middleware solve the issue. (Maybe nginx was not started correctly on my setup)

morucci:
@TristanCacqueray thanks for the review I've addressed your proposals.

TristanCacqueray (approved):

PR #886, opened 2022-06-05
Handle OIDC flow in backend

morucci:
This change implements the OIDC authentication code flow in the Monocle API.

Depends-on: https://github.com/change-metrics/dhall-monocle/pull/20

- /api/2/auth/login (where the User Agent is redirected after a Monocle login request). This endpoint returns HTTP 302 (location: to the OP login page).
- /api/2/auth/cb (where the User Agent is redirected after a successful login on the OP). This endpoint inspects the endpoint parameters (set by the OP) and generates a JWT via the local key on the Monocle API. This JWT contains information about the Monocle User (mUid, aliases, groups, ...) as well an expiration date.


.env.sample
```diff
@@ -11,3 +12,12 @@
...
 #MONOCLE_ELASTIC_ADDR=0.0.0.0
 #MONOCLE_ELASTIC_PORT=9200
+
+### Env vars to enable an OIDC Provider to authenticate Monocle users
+### The four first are manadatory to enable the OIDC Provider
+### MONOCLE_OIDC_USER_CLAIM will default to "sub" if not provided
+#MONOCLE_OIDC_ISSUER_URL=https://accounts.google.com
+#MONOCLE_OIDC_CLIENT_ID=12345
+#MONOCLE_OIDC_CLIENT_SECRET=superSecret
+#MONOCLE_OIDC_APP_PUBLIC_URL=http://localhost:8080
```
TristanCacqueray, line 22: Why not using the existing `MONOCLE_PUBLIC_URL`?

codegen/MonocleCodegen.hs
```diff
@@ -168,9 +172,13 @@ protoToServant pb =
...
     api =
       [ "type MonocleAPI =",
-        Text.intercalate "\n :<|>" $ map mkApi methods
+        Text.intercalate "\n :<|>" $ (map mkApi methods ) <> otherEndpoints
       ]
       where
+        otherEndpoints = [
+          "\"auth\" :> \"login\" :> Get '[JSON] NoContent",
+          "\"auth\" :> \"cb\" :> QueryParam \"error\" Text :> QueryParam \"code\" Text :> QueryParam \"state\" Text :> Get '[HTML] LoginInUser"
+          ]
```
TristanCacqueray, line 181: The codegen shouldn't have to deal with non-protobuf endpoint. Instead we could add the extra endpoints in the Monocle.Main module.

haskell/src/Monocle/Api/Jwt.hs
```diff
@@ -1,67 +1,53 @@
...
-    fromKeyMaterial,
-    genJWK,
-  )
-import Crypto.JWT qualified as Jose
-import Data.Aeson (decode)
+import Crypto.JWT (Error, JWK, KeyMaterialGenParam (RSAGenParam), genJWK)
 import Data.ByteString.Lazy qualified as BSL
+import Data.Map.Strict qualified as HM
+import Monocle.Config (OIDCProvider (..))
 import Monocle.Prelude
```
TristanCacqueray, line 18: Could we import the whole prelude instead?

haskell/src/Monocle/Config.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE ViewPatterns #-}
```
TristanCacqueray, line 5: The extension could be added to the cabal file too.

haskell/src/Monocle/Prelude.hs
```diff
@@ -553,3 +559,26 @@ data AnyJSON = AnyJSON
 
 instance FromJSON AnyJSON where
   parseJSON _ = pure AnyJSON
+
+genRandomBS :: IO ByteString
```
TristanCacqueray, line 563: That function looks suspicious, but I guess it works and we could rewrite it later.

TristanCacqueray (changes requested):

haskell/src/Monocle/Prelude.hs
morucci: I wrote something much simpler.

PR #894, opened 2022-06-17
Improve env var handling

morucci:
- Expose all env var in the CLI
- Prefix all env vars with MONOCLE_
- Prefix all compose related env vars to COMPOSE_MONOCLE_
- Complete env.sample
- Complete readme


CONTRIBUTING.md
```diff
@@ -86,7 +86,7 @@ export $(cat .secrets)
 cd haskell
 cabal repl monocle
 λ> import Monocle.Main
-λ> run 9879 "http://localhost:9200" "../etc/config.yaml"
+λ> run $ defaultApiConfig 9879 "http://localhost:9200" "../etc/config.yaml"
```
TristanCacqueray, line 89: nit: it might be good to avoid such breaking change, e.g. by keeping the existing run definition, and adding a new entrypoint which takes the ApiConfig.

Though I now wonder what's the point of these instructions... Perhaps it would be better to use the command line, and only provides such instructions for `ghcid`?

TristanCacqueray (approved):

CONTRIBUTING.md
morucci: I'm not sure that keeping the legacy just for the developer repl experience is worth the price of multiplying function definitions.

This provides an entrypoint for the developer to start component, reload after changes, ... but that's true a ghcid command to handle the reload might be great as well.

PR #895, opened 2022-06-19
Remove turtle pinning as proto3-wire got a new revision

morucci:
A new revision on hackage has been published that should
enable us to remove the constraint.


morucci:
The revision was only for turtle pinning but the bump of the cap of proto3-wire was not proposed as a revision. I've asked for it.
https://hackage.haskell.org/package/proto3-suite-0.5.0/revisions/

Moving as a Draft for the moment.

TristanCacqueray (approved):

Issue #847, opened 2022-03-23
Generalize Metrics to include self documentation


TristanCacqueray:
The work started with #792, and here are a list of things to do to complete the work:

- [x] Break all-in-one queries to use individual metrics. For example, `1st review mean time` is currently hard-coded in the getReviewStats query here: https://github.com/change-metrics/monocle/blob/f05cd10244e4e83f49906088150a486b5e60d98d/haskell/src/Monocle/Backend/Queries.hs#L1028-L1032 . Instead the metric should be defined with the new Metric data type.
- [x] Improve the Metric data type to include a long description, e.g. how to interpret the value.
- [x] Make the web interface get the query list and include tooltip on the graph with each value details
- [x] Enable direct metric query, e.g. `curl /api/metric/get/timeToMerge?q=...`
- [x] Enable trend query, e.g. `curl /api/metric/trend/timeToMerge?q=...&period=2weeks` which should output an histogram.

PR #901, opened 2022-07-05
More metrics under Metrics system

morucci:
Part of the implementation of #847

This change includes 4 more simple metrics into the new Metrics system.
Furthermore during the migration of the self merged metric I've figured out that the metrics must rely on the Change Merged Event and not on the Change object. Thus this changes brings a fix for this and a schema migration to add self-merged attribute to ChangeMergedEvent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 ### Removed
 ### Fixed
 
+- [api] The self-merged count on the Activity page was incorrect and it is now properly calculated.
+
 ## [1.6.0] - 2022-06-20
 
 ### Added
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 ### Removed
 ### Fixed
 
+- [api] Wrong count of self-merged changes metrics on Activity page.
```
TristanCacqueray, line 11:
```suggestion
- [api] The self-merged count on the Activity page was incorrect and it is now properly calculated.
```

TristanCacqueray (approved):

CHANGELOG.md
morucci: Thanks ! done.

PR #902, opened 2022-07-06
Add support for local URI redirect after Authentication

morucci:
- [x] backend support
- [ ] webui support


haskell/src/Monocle/Main.hs
```diff
@@ -35,7 +35,7 @@ type MonocleAPI' = MonocleAPI :<|> AuthAPI
 type RootAPI = "api" :> "2" :> MonocleAPI' :<|> MonocleAPI'
 
 type AuthAPI =
-  "auth" :> "login" :> Get '[JSON] NoContent
+  "auth" :> "login" :> QueryParam "uri" Text :> Get '[JSON] NoContent
```
TristanCacqueray, line 38: To avoid uri encoding issue, we can also use [`CaptureAll`](https://hackage.haskell.org/package/servant-0.19/docs/Servant-API.html#t:CaptureAll)

morucci: I don't think CaptureAll fills well the need as it capture path segments however it seems a simple parameter in the query string is well adapted to the need.

PR #910, opened 2022-07-20
Handle Metric trend query interval

morucci:
Related to  https://github.com/change-metrics/monocle/issues/847

### DIFF
--- a/schemas/monocle/protob/metric.proto
+++ b/schemas/monocle/protob/metric.proto
@@ -22,7 +22,7 @@ message ListResponse {
 }
 
 message Trend {
-  uint32 interval = 1;
+  string interval = 1;
 }
 
 message GetRequest {
--- a/src/Monocle/Search/Query.hs
+++ b/src/Monocle/Search/Query.hs
@@ -28,7 +28,8 @@ where
 
 import Control.Monad.Trans.Except (Except, runExcept, throwE)
 import Data.List (lookup)
-import Data.Time.Clock (secondsToNominalDiffTime)
+import Data.Time.Calendar (addDays, addGregorianMonthsClip, addGregorianYearsClip)
+import Data.Time.Clock (UTCTime (UTCTime), secondsToNominalDiffTime)
 import Data.Time.Format (defaultTimeLocale, formatTime, parseTimeM)
 import Database.Bloodhound qualified as BH
 import Monocle.Config qualified as Config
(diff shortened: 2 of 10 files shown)


morucci:
I'm wondering about the granularity as followup. Should we try to have:
```
data HistoInterval = Hour Natural | Day Natural | Week Natural | Month Natural | Year Natural deriving (Eq, Show)
```
Or less flexible:
```
data HistoInterval = Hour | 6Hours |  Day | 2Days | Week | 2Weeks | 3Weeks | Month  | 2Months | 3Months | 6Months |  Year | 2Years | 5Years deriving (Eq, Show)
```

What do you think @TristanCacqueray ?

src/Monocle/Search/Query.hs
```diff
@@ -162,11 +161,13 @@ queryFieldToDocument name = do
 -- | Resolves the actual document field for a given flavor
 getFlavoredField :: QueryFlavor -> Field -> Maybe Field
 getFlavoredField QueryFlavor {..} field
-  | field `elem` ["author", "group"] = Just $ case qfAuthor of
-    Author -> "author"
-    OnAuthor -> "on_author"
+  | field `elem` ["author", "group"] = Just author
```
morucci, line 165: Yes that's not related ... not the first time I have the issue but the "guards" + case related statements are not formatted the same way with ormolu via vscode and ormolu in ci. Not sure if vscode passes some parameter to ormolu because I only have one ormolu available in the path. Or maybe the hls provides ormolu out of the box ?
So here that's why there is a slight refactor to satisfied ormolu.

TristanCacqueray: I think https://github.com/change-metrics/monocle/pull/916 should fix that.

schemas/monocle/protob/metric.proto
```diff
@@ -22,7 +22,7 @@ message ListResponse {
 }
 
 message Trend {
-  uint32 interval = 1;
```
TristanCacqueray, line 25: I don't think using a protobuf enum to serialize trend interval is much better than encoding the data using a string. If we were using [typical](https://github.com/stepchowfun/typical#typical-data-interchange-with-algebraic-data-types), then perhaps it would be more practical to use a more correct representation, such as a sum type for the interval value.

In anycase, it may be easier to implement a custom string encoder/decoder to convert the user input form to the correct value, than directly providing the data type required by the backend.

src/Monocle/Search/Query.hs
```diff
@@ -222,6 +245,9 @@ parseRelativeDateValue now txt = relTimeToUTCTime <$> readMaybe (from txt)
               Hour -> hour
               Day -> day
               Week -> day * 7
+              -- TODO Use calender aware computation
```
TristanCacqueray: How does it work for elasticsearch, is there a feature to adapt the bucket size to match the right time range? Or should we use average value, e.g. 30.4 day per month, and 365.25 per year.

Perhaps we should pad the query, for example a monthly interval range could be clamped to start at the beginning of the month.

TristanCacqueray (approved):

schemas/monocle/protob/metric.proto
morucci: Ok so I set this as resolve as we use the TimeRange parser.

src/Monocle/Search/Query.hs
morucci: We use calendar_interval for Elasticsearch histogram. Elasticsearch interval start at the beginning of the interval period (from the elastic doc: "For example, if the interval is a calendar day, 2020-01-03T07:00:01Z is rounded to 2020-01-03T00:00:00Z."). [1]

Also we use "extended_bound" with min and max parameter which is the start and end date of the range. This make Elasticsearch to stick to the period and created the maximum number of buckets possible according the chosen interval period. [2]

[1]: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#search-aggregations-bucket-datehistogram-aggregation
[2]: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-histogram-aggregation.html#search-aggregations-bucket-histogram-aggregation-hard-bounds

PR #920, opened 2022-07-27
Move some Top style metrics into new metric system

morucci:
The effort is part of https://github.com/change-metrics/monocle/issues/847

Move some Top metrics under the new metric system.

### DIFF
--- a/src/Monocle/Api/Server.hs
+++ b/src/Monocle/Api/Server.hs
@@ -565,7 +565,8 @@ searchQuery auth request = checkAuth auth response
             SearchPB.QueryResponse . Just . SearchPB.QueryResponseResultActivityStats
               <$> Q.getActivityStats
           SearchPB.QueryRequest_QueryTypeQUERY_TOP_AUTHORS_CHANGES_COMMENTED ->
-            handleTopAuthorsQ queryRequestLimit Q.getMostActiveAuthorByChangeCommented
+            termsCountWord32ToResult . fromJust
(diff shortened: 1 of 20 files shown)


src/Monocle/Api/Server.hs
```diff
@@ -716,6 +729,31 @@ instance TrendPB Word32 where
...
       $ Q.toPBHistoInt <$> v
 
+class Num a => TopPB a where
+  toTopResult :: Maybe (Q.TermsCount a) -> MetricPB.GetResponse
+
+toTopResultNotSupported :: MetricPB.GetResponse
+toTopResultNotSupported =
+  MetricPB.GetResponse
+    . Just
+    $ MetricPB.GetResponseResultError "This metric does not support Top option"
```
TristanCacqueray, line 739: That's a bit odd, shouldn't it be:
```suggestion
  toTopResult :: Q.TermsCount a -> MetricPB.GetResponse
  
toTopResultOrFail :: Maybe (Q.TermsCount a) -> MetricPB.GetResponse
toTopResultOrFail = \case
  Just tc -> toTopResult tc
  Nothing -> MetricPB.GetResponse
    . Just
    $ MetricPB.GetResponseResultError "This metric does not support Top option"
```

And then in metricGet, use `toTopResultOrFail`. That way we don't have to handle the missing result in each instance.

morucci:
Rebased after the fourmolu landing.
Perhaps we can merge this ?

TristanCacqueray (approved):

PR #922, opened 2022-08-04
Fix elasticsearch modules cannot be a link

morucci:
Probably this no longer work with the previous version bump
of Elasticsearch.

Also this change makes elasticsearch-start script to create the
EL_HOME in ~/ instead of /var/lib that requires sudo.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -208,8 +208,8 @@ let
   mkHome = name: ''
     mkdir -p ${name} 2> /dev/null || {
       echo "${name}: creating"
-      /bin/sudo mkdir -m 0700 ${name}
-      /bin/sudo chown $(id -u) ${name}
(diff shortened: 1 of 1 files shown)


nix/default.nix
```diff
@@ -228,7 +228,7 @@ in rec {
       sha256 = "ocz3CJFf+diThmocrgSnhWW/fjuRLLyCxwUKl3Cm7WA=";
     };
   });
-  elasticsearch-home = "/var/lib/elasticsearch";
+  elasticsearch-home = "~/monocle-elasticsearch";
```
TristanCacqueray, line 231: Perhaps we should use xdg path, e.g. `~/.local/share/monocle/elasticsearch-home`

TristanCacqueray (approved):

PR #924, opened 2022-08-08
Integrate hspkgs overlay for haskell dependency

TristanCacqueray:
This change replace the local package set with a re-usable set
defined in podenv/hspkgs.

This change also bumps ghc to 9.2.4

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -462,10 +462,8 @@ in rec {
   fast-ci-run = mkRun "fast-ci" fast-ci-commands;
 
   reformat-run = mkRun "reformat" ''
-    # This needs apply-refact, but it doesn't build in our package set and that would requires pulling an extra ghc.
-    # Let's try again next time we bump nixpkgs.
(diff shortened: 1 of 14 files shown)


nix/default.nix
```diff
@@ -1,194 +1,76 @@
...
-              ref = "main";
-              rev = "fcb32257a37e34a48f70ddd55424394de4bb025c";
-            };
-          });
-
-          monocle = (hpPrev.callCabal2nix "monocle" self { }).overrideAttrs
-            (_: { MONOCLE_COMMIT = builtins.getEnv "MONOCLE_COMMIT"; });
-        };
-      };
+  # Add monocle and patch broken dependency to the haskell package set
```
TristanCacqueray, line 16: Note that even though we are using hspkgs, we can still override packages when needed, for example to pin bloodhound and morpheus to an older version. Though I'd be happy to move the overrides to hspkgs upstream when it makes sense.

morucci (approved):

PR #927, opened 2022-08-10
Add 'ttm' field to the query language

morucci:


src/Lentille.hs
```diff
@@ -203,3 +206,6 @@ isChangeTooOld date (change, _) =
   case changeUpdatedAt change of
     Just changeDate -> T.toUTCTime changeDate < date
     _ -> True
+
+swapDuration :: ChangeOptionalDuration -> ChangeEventOptionalDuration
```
TristanCacqueray, line 210: Perhaps a `From ChangeOptionalDuration ChangeEventOptionalDuration` instance would be more appropriate?

src/Monocle/Backend/Index.hs
```diff
@@ -315,11 +315,41 @@ upgradeConfigV3 = do
...
+    mkChangeBulkUpdate :: BH.IndexName -> EChange -> BulkOperation
+    mkChangeBulkUpdate indexName change =
+      BulkUpdate indexName (getChangeDocId change) $ toJSON change
+
 upgrades :: [(ConfigVersion, QueryM ())]
 upgrades =
   [ (ConfigVersion 1, upgradeConfigV1),
     (ConfigVersion 2, upgradeConfigV2),
-    (ConfigVersion 3, void upgradeConfigV3)
+    (ConfigVersion 3, void upgradeConfigV3),
```
TristanCacqueray, line 351: nit: we might want to use the fourmolu formater to avoid these un-necessary diff (by using prefix comma, like in dhall)

src/Monocle/Search/Query.hs
```diff
@@ -147,6 +147,7 @@ fields =
     ("group", (fieldText, flavoredField, "Group name")),
     ("branch", (fieldRegex, "target_branch", "Change branch name")),
     ("approval", (fieldText, "approval", "Change approval name")),
+    ("ttm", (fieldNumber, "duration", "Change duration (Time To Merge) in second(s)")),
```
TristanCacqueray, line 150: We might want to add a couple of test in https://github.com/change-metrics/monocle/blob/833d72ec8a4da358581c3624661ab4808392d3a1/src/Tests.hs#L459 to check the resulting elastic query.

TristanCacqueray (approved):

src/Lentille.hs
morucci: Both type are from codegen, and a From instance will raise a orphan instance warning.

src/Monocle/Backend/Index.hs
morucci: Yes probably in another patch.

src/Monocle/Search/Query.hs
morucci: Done

PR #931, opened 2022-08-15
Metric UI

morucci:


src/Monocle/Backend/Queries.hs
```diff
@@ -1278,9 +1277,8 @@ metricChangesMerged = changeEventCount mi EChangeMergedEvent
       "changes_merged"
       "Changes merged count"
       "The count of changes merged"
-      ( Just $
-          "The metric is the count change merged events."
-            <> changeEventFlavorDesc
+      ( "The metric is the count change merged events."
+          <> changeEventFlavorDesc
```
TristanCacqueray, line 1281: We might want to have a look at string interpolation library to make our life easier. Perhaps https://hackage.haskell.org/package/string-interpolate or even https://hackage.haskell.org/package/PyF

morucci: I've add the use of string-interpolate. +1

src/Monocle/Api/Server.hs
```diff
@@ -695,10 +695,27 @@ metricList auth _request = checkAuth auth . const $ response
     MetricPB.MetricInfo
       { metricInfoName = from miName
       , metricInfoDescription = from miDesc
-      , metricInfoLongDescription = mempty
+      , metricInfoLongDescription = from miLongDesc
       , metricInfoMetric = from miMetricName
       }
 
+metricInfo :: AuthResult AuthenticatedUser -> MetricPB.InfoRequest -> AppM MetricPB.InfoResponse
```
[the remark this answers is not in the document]
morucci: Yes, if you start the UI, you'll see that is convenient to get the long description on each metric page.

PR #942, opened 2022-09-14
Refactor crawler metadata representation

TristanCacqueray:


morucci (changes requested):
getCrawlerMetadataDocId needs to follow the same behavior as before. Other than that that seems good at first glance !

morucci (approved):

PR #943, opened 2022-09-15
Introduce structured logging facility

TristanCacqueray:
This change enables migrating away from the LogEvent sum type into a structured logging system.
The goal is to simplify the logging implementation. It is presently tricky to correlate log message with their location in the code, and adding richer information such as Entity is painful, see the entity refactor.

### DIFF
--- a/src/Monocle/Logging.hs
+++ b/src/Monocle/Logging.hs
@@ -1,4 +1,10 @@
 -- | Monocle log events
+-- Note: [Monocle Structured Logging]
+--
+-- Structure comes from the 'Data.Aeson.Series'. They can be created using
+--    key .= value
+-- And combined using mappend:
+--    key1 .= value1 <> key2 .= value2
 module Monocle.Logging (
   -- * Logging effect
   HasLogger (..),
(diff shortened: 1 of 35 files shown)


src/Monocle/Logging.hs
```diff
@@ -62,17 +63,16 @@ data LogEvent
...
-  | RefreshIndex Config.Index
   | OIDCCallbackCall (Maybe Text) (Maybe Text) (Maybe Text)
   | OIDCProviderTokenRequested Text
-  | JWTCreated Text Text
-  | JWTCreateFailed Text Text
+  | LogStructured Text [Aeson.Pair]
 
 instance From LogEvent Text where
   from = \case
+    LogStructured msg attrs
```
morucci, line 73: I believe that info like "System ready" which is passed via msg here and the LogAuthor must be mandatory fields of the LogStructured type. It seems that the intend is to make logs processable via a program then each useful info must be part of the JSON struct.

TristanCacqueray: Well the current SystemReady event doesn't have a LogAuthor. I think what we want instead is the source location of the message, e.g. `Monocle.Api`, `Macroscope.Worker` or `Lentille.GitHub`, and we will get that automically by using [HasCallStack](https://hackage.haskell.org/package/base-4.13.0.0/docs/GHC-Stack.html#t:HasCallStack).

For crawler information, these should be added as attrs, e.g. `["crawler" .= crawlerName, "entity" .= entity]`

morucci (approved):
Great well done ! I walked over the code changes and that looks good. I did a quick test and I see the structured logs as well as location (module + line).

- There's still a hlint issue
- Do you think we could completely get ride of LogCrawlerContext ?
- Maybe add changelog entry to tell that the api and crawler log format has been changed to JSON structure logging ?

For the BlockArguments extension I'm fine with it.

PR #954, opened 2022-09-20
Introduce the effectful library

TristanCacqueray:
The goal of this PR is to replace the ReaderT over IO and the mtl style constraints with a simpler system.
This change introduces the Monocle.Effects module. The module implements re-usable static effects (ElasticEffect and MonoQuery) to remove AppM and QueryM from the api code.

The migration will be completed after:
- [x] Replacing LentilleM with new effects for the crawler code.
- [x] Removing mtl from the Monocle.Prelude.
- [ ] Demonstrating that adding new effects, such as tracing and reporting, is actually simple.


src/Monocle/Servant/HTTP.hs
```diff
@@ -40,7 +44,7 @@ type MonocleAPI =
     :<|> "crawler" :> "add" :> Auth '[JWT, Cookie] AuthenticatedUser :> ReqBody '[JSON] Monocle.Protob.Crawler.AddDocRequest :> Post '[PBJSON, JSON] Monocle.Protob.Crawler.AddDocResponse
     :<|> "crawler" :> "commit" :> Auth '[JWT, Cookie] AuthenticatedUser :> ReqBody '[JSON] Monocle.Protob.Crawler.CommitRequest :> Post '[PBJSON, JSON] Monocle.Protob.Crawler.CommitResponse
     :<|> "crawler" :> "get_commit_info" :> Auth '[JWT, Cookie] AuthenticatedUser :> ReqBody '[JSON] Monocle.Protob.Crawler.CommitInfoRequest :> Post '[PBJSON, JSON] Monocle.Protob.Crawler.CommitInfoResponse
-server :: ServerT MonocleAPI AppM
+server :: ApiEffects es => ServerT MonocleAPI (Eff es)
```
morucci, line 47: This will need an update in the codegen

src/Monocle/Api/Test.hs
```diff
@@ -26,20 +38,40 @@ mkAppEnv workspace = withLogger \glLogger -> do
       aEnv = Env {..}
   pure $ AppEnv {..}
 
-withTestApi :: IO AppEnv -> (Logger -> MonocleClient -> Assertion) -> IO ()
+--  Note: when running Effect, the order is set
+runAppEnv :: AppEnv -> Eff (ElasticEffect : MonoConfigEffect : E.Reader AppEnv : LoggerEffect : E.Error ServerError : E.Fail : E.Concurrent : IOE : '[]) a -> IO a
```
blackheaven, line 42:
```suggestion
runAppEnv :: AppEnv -> Eff ('[ElasticEffect, MonoConfigEffect, E.Reader AppEnv, LoggerEffect, E.Error ServerError, E.Fail, E.Concurrent, IOE]) a -> IO a
```

TristanCacqueray: I don't quite get why sometime the Effects needs to be listed in order, and sometime a constraint is enough, hence the weird signature here and there. Thank you for the suggestion!

blackheaven: When you use the effects, order does not matter (as long as the effects are listed in the constraints).

However, whenever you interpret them, you have to "prove" that all the effets are interpreted and to do so, you need a concrete type.

I don't know for `effectful`, but for `polysemy`, you can put an interpreter (or an interceptor) anywhere in the code, and all the effects will be "removed" during your "final" interpreter run, without breaking your code.
It works because your effect library consider that, your type not being fully define, an "additional" (duplicated) effect was present, and so, it removes it.

TristanCacqueray: I see thank you very much for confirming. I guess the order is necessary anywhere Eff is on the left side of the arrow and IO is on the right side. For example, this `withTenant` doesn't interpret the effect locally, but it's like contaminated by the call to `withTenantConfig` below:

https://github.com/change-metrics/monocle/blob/0b806b2beaacf47d0dff03f909b2a179320d8f78/src/Monocle/Backend/Test.hs#L88-L111

TristanCacqueray:
The PR is almost finished, its missing:

- [ ] Update builder image (or replace container build with nix)
- [ ] Adapt the codegen
- [ ] Move the [Monocle.Effects](https://github.com/TristanCacqueray/monocle/blob/mono-effects/src/Monocle/Effects.hs) note to an ADR to explains the design and how it works.

src/Monocle/Api/Test.hs
blackheaven: > I guess the order is necessary anywhere Eff is on the left side of the arrow and IO is on the right side

Usually yes, but it can change with hofs:

```queryAsync :: (ElasticEffect :> es, MonadIO m) => (Eff es a :> m a) -> Eff es a -> Eff es a```

TristanCacqueray:
With 4161cd7 , I think this PR is completed, the ReaderT and their associated Monad are now removed. Note that this change is documented in this blog post: https://www.softwarefactory-project.io/introducing-an-effects-system-for-monocle.html . The PR adds about 500 new lines, which can be explained by:
- Three new Effectful generic modules are included (Servant, Env and Prometheus) and they should be published as individual libraries.
- The Monocle.Effects module includes new tests and documentations.
- The change aimed for a drop-in replacement, essentially keeping the existing error handling and http retry logic. Refactoring these implementation will yield better result in follow-ups.

PR #955, opened 2022-09-27
Build the container using nix docker tools

TristanCacqueray:
This change removes the builder container so that Monocle uses the same toolchain for development and production.


.github/workflows/docker.yaml
```diff
@@ -38,9 +45,13 @@ jobs:
           CRAWLERS_API_KEY=secret
           GITHUB_TOKEN=123
           EOF
+      - name: Build monocle-exe container
+        run: "nix build .#containerMonocle; docker load < ./result"
+      - name: Build docker image
+        run: |
```
morucci, line 51: It seems we issue twice the same command

morucci (changes requested):

.github/workflows/docker.yaml
TristanCacqueray: What do you mean? The first docker load setup the "monocle-exe" image, and docker build produces the final "monocle" image, with the web assets.

morucci: https://github.com/change-metrics/monocle/blob/6099cfa475588a6497f90e3f952ad710949bee6c/.github/workflows/docker.yaml#L50-L55

TristanCacqueray: oh ok, then 8f9ca41 fix that :)

morucci:
@TristanCacqueray I did a build image with nix, then run docker build and the devel demo deployment was a success.
So Let's merge \o/

morucci:
Need to check if the master image is well published by the CI

PR #956, opened 2022-09-28
Fix monocle unable to find CA bundle

morucci:
Since the recent change to use nix to build the monocle container image, monocle was unable to find the ca-bundle.crt file.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -480,12 +480,17 @@ in rec {
     # Create a passwd entry so that openssh can find the .ssh config
     createPasswd = "echo ${user}:x:0:0:monocle:/${home}:/bin/bash > etc/passwd";
 
+    # Make ca-bundles.crt available to HSOpenSSL as plain file
+    # https://hackage.haskell.org/package/HsOpenSSL-x509-system-0.1.0.4/docs/src/OpenSSL.X509.SystemStore.Unix.html#contextLoadSystemCerts
+    fixCABundle =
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -42,4 +42,9 @@ RUN sed -e 's|-bs-no-version-header|-bs-no-version-header", "-warn-error -a+5+6+
 ################################################################################
 FROM quay.io/change-metrics/monocle-exe:latest
 
+# https://hackage.haskell.org/package/HsOpenSSL-x509-system-0.1.0.4/docs/src/OpenSSL.X509.SystemStore.Unix.html#contextLoadSystemCerts
+# Nix cacert installs a link to the nix store in /etc/ssl/certs/ca-bundle.crt
+# Let's copy it in /etc/pki/tls/certs as a regular file
+RUN mkdir -p /etc/pki/tls/certs/ && cp /etc/ssl/certs/ca-bundle.crt /etc/pki/tls/certs/ca-bundle.crt
```
TristanCacqueray, line 48: I think it would be better to simply run that comment in https://github.com/change-metrics/monocle/blob/master/nix/default.nix#L481 . Just make the path relative:

  mkdir -p etc/pki/tls/certs && ln -s /etc/ssl/certs/ca-bundle.crt etc/pki/tls/certs/ca-bundle.crt

morucci: The `isRegularFile` here https://hackage.haskell.org/package/HsOpenSSL-x509-system-0.1.0.4/docs/src/OpenSSL.X509.SystemStore.Unix.html#contextLoadSystemCerts might cause trouble if we keep the link.

Alright I'll do the change within nix instead of inside the Dockerfile.

TristanCacqueray (approved):

PR #967, opened 2022-10-12
Web: Remove adding the Bearer header in the webapi functions

morucci:
The API is already capable of authenticating from the Cookie.

Server side we make sure that the auth/cb OIDC callback endpoint forces the browser to set the cookie.


src/Monocle/Api/Jwt.hs
```diff
@@ -83,16 +83,21 @@ data LoginInUser = LoginInUser
   }
   deriving (Show, Eq, Ord)
 
+-- TODO - Add the Set-Cookie Header "Set-Cookie: JWT-Cookie=...."
```
TristanCacqueray, line 86: Actually, that should be the only way to set the cookie, such http-only cookie can't be set with javascript i think. For example, using: https://hackage.haskell.org/package/servant-auth-server-0.4.7.0/docs/Servant-Auth-Server.html#v:acceptLogin

src/Monocle/Api/Jwt.hs
```diff
@@ -83,16 +83,21 @@ data LoginInUser = LoginInUser
...
+      H.title "Redirecting after a successful login ..."
     H.body do
       H.script
         ( H.toHtml
             ( "localStorage.setItem('api-key','"
                 <> liJWT
                 <> "');"
+                <> "document.cookie = \"JWT-Cookie="
+                <> liJWT
+                <> ";path=/\";"
```
TristanCacqueray, line 100: That should not be necessary.

src/Monocle/Api/Jwt.hs
morucci: Yes, agree, with that solution the flow is working however I'm not happy with it and I'll prefer to let servant to set the proper header for the cookie. I'll try with your proposal.

morucci:
I did the implementation based on https://github.com/haskell-servant/servant/tree/master/servant-auth. It seems to work fine, however there is an issue that happen randomly due to the X-XSRF token that it seems to be sent by axios too (I have X-XSRF-TOKEN header + XSRF-TOKEN in a cookie) and on some requests done by axios to the API both tokens are different then servant responds forbidden.

I guess it worked well before as the first auth mechanic on the servant API is [JWT, ...] and previously this one matched for the requests done by the webApp. Now we use Cookie as auth mechanic then the issue appeared.

I'll check how to deactivate that X-XSRF-TOKEN in axios.

morucci:
@TristanCacqueray Hi, I think that, now, the state of that PR is alright for review. Please see the last commit.

src/Monocle/Api/Server.hs
```diff
@@ -934,24 +967,37 @@ handleLoggedIn err codeM stateM = do
...
+      -- Here we create the JWT Session Cookie that will be used by the browser to authenticate requests
+      mApplyCookies <-
+        liftIO $
+          acceptLogin
+            -- The nested JWT owns the expiry
+            (cookieSettings {cookieExpires = Just expiry})
+            jwtCfg
+            authenticatedUser
+      -- Here we create a already expired JWT to be set as the Monocle Cookie. This cookie is used by web app
+      -- to extract non sensible authenticated user info
```
TristanCacqueray, line 986: Why not provide this information through the info endpoint instead? That sounds like a design issue, we shouldn't have to provide multiple cookies like that.

src/Monocle/Api/Server.hs
```diff
@@ -905,16 +917,37 @@ handleLogin uriM = do
     loc <- O.prepareAuthenticationRequestUrl (mkSessionStore oidcenv Nothing uriM) oidc [O.openId] mempty
     return (show loc)
 
+-- This session cookie is made for the Monocle Web APP to figure out the authenticated user context
+makeMonocleCookie :: LBS.ByteString -> SetCookie
+makeMonocleCookie jwt = do
+  defaultSetCookie
+    { setCookieValue = LBS.toStrict jwt
+    , setCookieHttpOnly = False -- This make the cookie readable from the Web APP
```
TristanCacqueray, line 925: Is this set so that the username can be read by the webUI from the auth cookie? If yes, then we should keep the http-only, and use an extra query to get the info from a API response body.

src/Monocle/Api/Server.hs
```diff
@@ -905,16 +917,37 @@ handleLogin uriM = do
...
+    , setCookieName = "Monocle"
+    , setCookieSameSite = Just sameSiteLax
```
TristanCacqueray, line 927: Why not use Strict ?

src/Monocle/Main.hs
```diff
@@ -158,7 +158,8 @@ run' ApiConfig {..} aplogger = E.runConcurrent $ runLoggerEffect do
 
     let settings = Warp.setPort port $ Warp.setLogger aplogger Warp.defaultSettings
         jwtCfg = localJWTSettings
-        cfg = jwtCfg :. defaultCookieSettings :. EmptyContext
+        cookieCfg = defaultCookieSettings {cookieXsrfSetting = Nothing}
```
TristanCacqueray, line 161: I think that's ok to do since we don't have POST form or anything that could be maliciously manipulated, though it would be good to know why this is not working.

morucci: Agreed. My theory about the potential cause is explained in the commit message. Perhaps using session in Axios might solve the issue. I did some research to see if other are affected by that behavior but I found nothing relevant. I'll be happy to show you the debug console of the browser ... then if you get any idea about how we could fix this (though if it worth the try)

src/Monocle/Api/Server.hs
morucci: Sure I can change this.

src/Monocle/Api/Server.hs
morucci: Yes, this cookie is only for the Web APP to be able to read the authenticated user info. However this is **NOT** the auth cookie. This cookie is the "Monocle" cookie which contain an **invalidated** JWT containing the same information as the one in the local storage previously.

The auth cookie is the JWT-Cookie created by the acceptLogin function of servant-auth-servant. This one is httpOnly.

src/Monocle/Api/Server.hs
morucci: By changing the way the JWT is shared between the Web APP and the API I wanted to minimize the amount of changes especially inside the Web APP, by also keeping the current Web APP behavior, which is the case with that change. However I agree there are rooms for improvement like as you said relying only on the API to get the user's info and remove the additional cookie. I'll create a new followup story.

morucci:
Let me know if this is OK to land as it is ?

TristanCacqueray (approved):

PR #1000, opened 2022-11-28
Release 1.8.0

morucci:


morucci:
The failure in the k8s CI job is expected as the jobs expects to fine the 1.8.0 container image.

PR #1015, opened 2023-02-12
Remove prometheus and grafana from default shell

morucci:
Provide a monitoring shell: nix develop .#monitoriing

This saves disk space when entering a monocle dev shell. Those tools are extra for the specific monitoring use case.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -135,7 +135,7 @@ in rec {
                 - CRAWLER_TARGET
     '';
   };
-  prom-home = "/var/lib/prometheus";
+  prom-home = "~/.local/share/monocle/prometheus-home";
   promStart = pkgs.writeScriptBin "prometheus-start" ''
     ${headers}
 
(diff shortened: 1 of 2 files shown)


nix/default.nix
```diff
@@ -302,14 +302,13 @@ in rec {
...
-    monocleWebStart
-    monocleGhcid
-  ];
+  services-req =
+    [ elasticsearchStart monocleReplStart monocleWebStart monocleGhcid ];
+
+  monitoring-shell = hsPkgs.shellFor {
+    packages = p: [ p.monocle ];
+    buildInputs = [ promStart grafanaStart ];
+  };
```
TristanCacqueray, line 311: This does not need hsPkgs, it would be better to use `mkShell` directly:

```suggestion
  monitoring-shell = pkgs.mkShell {
    buildInputs = [ promStart grafanaStart ];
  };
```

But then, now that there is a flake, it would be even better to use an app entry like: https://github.com/change-metrics/monocle/blob/7a6fb4c3c3e30d85d375b1f7f10a57e5b2d78a71/flake.nix#L46-L49
 by setting the `program` attribute to the `promStart` and `grafanaStart` value. That way we could simply do `nix run .#prometheus` and `nix run .#grafana`.

morucci: Good idea! see the followup patch.

TristanCacqueray (approved):

PR #1026, opened 2023-04-22
Bump hspkgs for GHC-9.6.1

TristanCacqueray:


TristanCacqueray:
hlint and fourmolu are conflicting about the constrains paren :)

morucci:
Thanks ! It seems that the CI is failing due to some hlint issues.

src/Monocle/Backend/Test.hs
```diff
@@ -1119,8 +1119,8 @@ testTaskDataAdd = withTenant doTest
     assertEqual'
       "Check adding matching taskData"
       [ ("44", Nothing)
-      , ("43", Just [I.toETaskData crawlerName td43])
       , ("42", Just [I.toETaskData crawlerName td42])
+      , ("43", Just [I.toETaskData crawlerName td43])
```
TristanCacqueray, line 1123: not sure if the order matter here.

morucci (approved):

PR #1027, opened 2023-05-06
Add a Lentille crawler for fetch Github PRs of a guiven user

morucci:


TristanCacqueray (approved):

TristanCacqueray:
Well done! I guess we need to update the crawler configuration? Or do we want to run such crawler dynamically, e.g. through a new api endpoint?

morucci:
> Well done! I guess we need to update the crawler configuration? Or do we want to run such crawler dynamically, e.g. through a new api endpoint?

I think having a crawler config might be convenient as will be valuable out of the box by integrating with the UI and macroscope.

I haven't thought about the dynamic crawler via l'api http ... what could the usage ?

TristanCacqueray:
> I haven't thought about the dynamic crawler via l'api http ... what could the usage ?

It would be nice to use this new lentille without having to update the config. Perhaps the api could start a new background thread to fetch a given user activity?

morucci:
> > I haven't thought about the dynamic crawler via l'api http ... what could the usage ?
> 
> It would be nice to use this new lentille without having to update the config. Perhaps the api could start a new background thread to fetch a given user activity?

Yes that's a good idea, and that could also makes the CLI (or other external consumer) to only relies on the API to run a stream.

Is that lentille background process will also stores fetched data in the database or just behaves like the actual CLI commands (JSON dump) ?

TristanCacqueray:
Both would be nice! The client would get the dump and the data could be indexed to be visualized in the UI.

PR #1033, opened 2023-05-10
conf: add option for the crawler loop delay

TristanCacqueray:


TristanCacqueray:
Perhaps this should be stored in a new CrawlerConfig schema, what do you think @morucci ? Also why isn't the CI running?

morucci (approved):

Issue #1044, opened 2023-05-25
Configure Log Level


SaschaLucius:
Hey,
I'm using monocle with kubernetes in an AWS Cloud environment.
Since my load balancer is checking the health of the application frequently, I'm facing a lot of info log messages of the following kind:
` - - [25/May/2023:09:24:44 +0000] "GET /health HTTP/1.1" 200 - "" "ELB-HealthChecker/2.0"`

Is there any way to configure the log level of the server monocle is using?

Have looked into the code, and I'm not sure how to do it.
Maybe https://github.com/change-metrics/monocle/blob/master/k8s/api-service.yaml could be the entrance point.

TristanCacqueray:
Hi, logging is presently not configurable, and I'm not sure what should be the options. Though I think it's ok to silence `/health` and `/metrics` request, as proposed in #1045 .

SaschaLucius:
@TristanCacqueray thank you for your fast response and a proposed solution.
My initial idea was to change the log level to warning instead of info to ignore these status messages.
But your solution would also help to solve my problem.

Therefore, thank you
I will have a sort look at your PR, but my Haskell skills are a little bit rusty

TristanCacqueray:
You are welcome, and thank you for the feedback. I meant, having configurable log level sounds like a good feature to me, but we need to define how this would work. For example, HTTP logs are managed by the http server library (warp), and they don't have a severity attached. The monocle's backend code do use  warning/info/debug levels. So I guess we could have:

- Drop message below a global level setting, e.g. an env variable: MONOCLE_LOG=warning|info|debug.
- Handle configuration per components, e.g. MONOCLE_HTTP_LOG=... , MONOCLE_API_LOG=... 

We also talked about pushing the logs in elasticsearch too, that way the console output could only have warnings.

SaschaLucius:
My question would be, are we already discussing premature optimization?

Your proposed fix helps me with my issue, so I'm happy.

E.g. the proposed fix for Nginx is to ignore dependent on the agent https://ismailyenigul.medium.com/disable-logging-on-nginx-pod-for-elb-healthchecker-2-0-user-agent-e745ad55a7b4 .

Future wise, I agree a global level setting or a component wise would be nice, but until someone else than me is complaining about it, I would not invest into it ;)

PR #1045, opened 2023-05-25
api - disable /health and /metrics request logging

TristanCacqueray:
Related #1044


SaschaLucius (commented):
Looks good from my side.
I'm just thinking about if someone cares about non 200 responses from /health or /metrics in the logs.
At least I don't care.

TristanCacqueray:
Good point, last commit checks for non 200 responses.

morucci (approved):

PR #1050, opened 2023-08-16
Metrics - peers strength - add some explanation about the computation

morucci:


src/Monocle/Backend/Queries.hs
```diff
@@ -714,14 +717,22 @@ getAndSortToPeerStrength limit authors_peers =
      in
       toPSR <$> peers_authors
 
+-- This function is used by the API peersStrength metrics (QUERY_TOP_AUTHORS_PEERS)
```
TristanCacqueray, line 720: You might want to prefix document comment with `-- |` so that they appears in haddock:

```suggestion
-- | This function is used by the API peersStrength metrics (QUERY_TOP_AUTHORS_PEERS)
```

src/Monocle/Backend/Queries.hs
```diff
@@ -673,9 +673,12 @@ getChangeEventsTop limit docs qfield qf =
...
 -- | peer strength authors
 data PeerStrengthResult = PeerStrengthResult
-  { psrAuthor :: Text
-  , psrPeer :: Text
-  , psrStrength :: Word32
+  { -- The author name
+    psrAuthor :: Text
+  , -- The peer reviewer name
+    psrPeer :: Text
+  , -- The count of reviews + comments events
```
TristanCacqueray, line 680: Record field documentation also needs to be adjusted:

```suggestion
  { -- | The author name
    psrAuthor :: Text
  , -- | The peer reviewer name
    psrPeer :: Text
  , -- | The count of reviews + comments events
```

PR #1052, opened 2023-08-19
[ui]: initial proof of concept interface using butler

TristanCacqueray:
This change implements the following integration between monocle and butler:

- [x] A sample app to perform query.
- [x] Butler's display websocket served along the monocle api.
- [ ] Run butler context using monocle logging system (or the otherway around).
- [ ] App storage using a custom elasticsearch index.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,8 +1,46 @@
 {
   "nodes": {
+    "butler": {
+      "inputs": {
+        "hspkgs": "hspkgs"
+      },
+      "locked": {
+        "lastModified": 1692469360,
+        "narHash": "sha256-wXbQuCM9g+pBDetpQl+NIsFyKNfAoc8ABJ8HETMe1TY=",
+        "owner": "TristanCacqueray",
+        "repo": "haskell-butler",
+        "rev": "94af3ff012940511dec2ac62acf830641d636134",
+        "type": "github"
+      },
+      "original": {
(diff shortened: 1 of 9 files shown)


TristanCacqueray:
@morucci here is a little poc to integrate butler. It's still missing proper logging and storage, but for now, if you go to "/view", then you'll get a butler app that makes a query :-)

morucci:
When I access /view route my browser keep trying again and again 'ws/htmx' but server side I see "59786ms error   <1>init Butler.Display.WebSocket:91     Unknown websocket session {"session": null}" at each connection attempt.

TristanCacqueray:
Do you have oidc auth enabled and you are not logged-in the regular ui?

morucci:
That was it (my OIDC config) now I see "Change count: 0" which is expected I guess as I don't have any data :)

morucci:
Then I'm wondering if, as a follow-up of this POC, that could be an opportunity to move the search author implementation (which is already via htmx) to the butler/ws ?

TristanCacqueray:
Well that's the reason I made this prototype, so that we can experiment with such UI implementation :) My goals here are two folds:

- Implement a grafana-like interface, where user can create and share dashboards composed of individual panels.
- Improve butler to smooth such integration, so that we could for example, use it for zuul-weeder as well.

PR #1053, opened 2023-08-29
add troubleshooting entry for network error message on web ui

yetanotheralex:
Was faced with the following error when trying to setup monocle on a remote host:
image

Ends-up I needed to set `MONOCLE_PUBLIC_URL` to my host name to make the queries succeed, but the error was a bit confusing, so I'm at least suggesting a troubleshooting entry for this.


TristanCacqueray (approved):
Thanks!

morucci (approved):

PR #1054, opened 2023-08-29
ci: remove the patched docker container

TristanCacqueray:
This change tries to fix the docker ci.


TristanCacqueray:
@morucci I'm fast tracking this change to fix the CI.

PR #1056, opened 2023-09-08
Update README for OIDC callback URI

avanier:
This mini-PR updates the documented callback URL to match the calls actually being performed.


morucci (approved):
Thanks !

PR #1059, opened 2023-09-16
release - prepare release 1.9.0

morucci:


morucci:
k8s job fails due to the use of the not yet published image.

Issue #1060, opened 2023-09-17
Github token expired situation


epaillous:
Hello there! 
First of all, thank you for this project : I've used it some months ago and was able to retrieve some data for my teams!
However, I've stopped using it during 2 months, so my github token expired. I've regenerate new one, and try to replace it in the .secrets but it seems that the crawler can not fetch data anymore, here is the error I get : 
```
 WARNING Lentille.GitHub.RateLimit:62: Unexpected error {"index":"my_index","crawler":"github-my-crawler","stream":"Changes","err":"\"FetchErrorNoResult\""}
```

I've tried to wipe all data from the crawler (even if I do not really want to do that because it took some times to retrieve it 😅 ) but the command failed : 
```
docker-compose run --rm --no-deps api monocle janitor wipe-crawler-data --elastic elastic:9200 --config /etc/monocle/config.yaml --workspace my-workspace --crawler-name my-crawler

Invalid argument `wipe-crawler-data'

Usage: monocle janitor COMMAND
  Maintain the database

```

Any ideas ? 😇

morucci:
Hi Emilie,

The `wipe-crawler-data` command is part of the last release 1.9.0 so be sure to bump your deployment to this release. See the blog post here for the procedure to follow: https://changemetrics.io/posts/2023-09-17-version-1.9.0-release.html

However to replace an expired token, usually you just have to:

- generate a new token (be sure to read the section about the old and new fine grained token https://github.com/change-metrics/monocle#change)
- stop the crawler process with: `docker-compose stop crawler`
- update the `.secrets` file
- start the crawler process with: `docker-compose start crawler`

morucci:
No the delay did not changed. However there is a new setting for configuring it if you want to change the default: https://github.com/change-metrics/monocle/pull/1033

epaillous:
Hi @morucci,
thanks for your quick answer! 
Not sure it works, I have added the loop_delay in my config.yaml like this : 
```
workspaces:
  - name: my-workspace
    crawlers:
      - name: github-my-crawler
        provider:
          github_organization: my-org
          github_repositories:
            - my-repo
        update_since: '2023-04-01'
        loop_delay_sec: 120
```
stop and start the crawler by doing : 
```
docker-compose stop crawler
docker-compose start crawler
```
but it seems that the crawler is still fetching only every 600 seconds 😢 
Moreover, despite the `update_since: '2023-04-01'` it seems that since I've regenerate my token it does not fetch old data, see this : 
[...]

PR #1062, opened 2023-09-19
Handle Authentication Error

TristanCacqueray:
This change avoids retrying requests that fail with a 401 error. #1060

### DIFF
--- a/src/Lentille/GitHub/RateLimit.hs
+++ b/src/Lentille/GitHub/RateLimit.hs
@@ -10,7 +10,7 @@ import Lentille.GitHub.Types
 import Lentille.GraphQL
 import Monocle.Prelude
 import Network.HTTP.Client (responseBody, responseStatus)
-import Network.HTTP.Types (Status, badGateway502, forbidden403, ok200)
+import Network.HTTP.Types (Status, badGateway502, forbidden403, ok200, unauthorized401)
 
 import Effectful.Retry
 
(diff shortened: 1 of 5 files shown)


TristanCacqueray:
@morucci I've replaced the streaming error handling with a simpler approach in this PR. Though I don't expect it to fix the issue, and if it does, it would be good to know why. What we observed is that for take datas, the processing ends correctly:

```
2023-09-19 17:07:53 WARNING Lentille.GraphQL:235: Could not fetch the current rate limit {"index":"monocle2","crawler":"gc-monocle","stream":"TaskDatas"}
2023-09-19 17:07:53 WARNING Macroscope.Worker:132: Error occured when consuming the document stream ...
```

But for projects and pull-requests, the error is somehow ignored:

```
2023-09-19 17:08:03 WARNING Lentille.GraphQL:235: Could not fetch the current rate limit {"index":"monocle2","crawler":"gc-monocle","stream":"Changes"}
2023-09-19 17:08:03 INFO    Macroscope.Worker:188: Continuing on next entity {"index":"monocle2","crawler":"gc-monocle","stream":"Changes"}
```

TristanCacqueray:
@morucci in the last commit I've also added a test to validate the logic, and the test pass with or without the refactor, so I think the issue is somewhere else in the GraphQL module.

morucci:
Thanks for the refactor. Let's merge it. However yes it is a weird that processStream do not receive the GraphError from streamFetch.

morucci (approved):

PR #1066, opened 2023-09-29
crawler: Fix target branch field not set on ChangeEvent Docs

### LINKED ISSUE #1065: Target branch in the ChangeMergedEvent
Hi 

why we do not have a target branch in the ChangeMergedEvent. target branch exists in Change event but this is not very handy :(
Can we add a target branch to ChangeMergedEvent?

morucci:
This change fixes an issue that happen when filtering by branch in the search query then resulting graphs are empty. This was because most of the graph related metrics computation relies on the ChangeEvent docs. Adding the missing field fixes that issue.

fix #1065

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 - [crawler] Correctly handle errors and do not commit new changes.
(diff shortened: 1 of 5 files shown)


TristanCacqueray (approved):

leonid-deriv:
Is a new docker image available or I need to build it my myself. I do not see changes in the log

PR #1070, opened 2023-10-05
container-image: provide a Monocle container based on UBI 8

morucci:


src/Monocle/Config.hs
```diff
@@ -502,7 +502,7 @@ mkIndexName name = do
   let check explanation p = if p then Right () else Left explanation
   check "Is empty" $ not $ T.null name
   check "Is longer than 255 bytes" $ BS.length (T.encodeUtf8 name) < 256
-  check "Contains uppercase letter(s)" $ T.all (\x -> not (isLetter x) || isLowerCase x) name
+  check "Contains uppercase letter(s)" $ T.all (\x -> not (isLetter x) || isLower x) name
```
blackheaven, line 505: What was the issue?

morucci: On that image I use ghc 9.4 and it seems that `isLowerCase` is only available from base 4.18 [1], however base 4.18 is available with ghc 9.6 [2].

This PR uses ghc 9.4, however ideally using 9.6 will be better as that's the one we use under nix. Dependencies issues went a bit too far using 9.6 then I stopped and fallback to use 9.4 for now.

[1]: https://hackage.haskell.org/package/base-4.18.1.0/docs/Data-Char.html#v:isLowerCase
[2]: https://wiki.haskell.org/Base_package

blackheaven: Thanks for the explanation.

TristanCacqueray (approved):

PR #1073, opened 2023-10-25
crawler - github - index the mergeCommit's sha

### LINKED ISSUE #1072: Add merge_commit_sha to Change event
Can we add merge_commit_sha to the Change event?

Currently, only code commits are included. We believe this should not be very difficult.
We need to correlate merge_commits to the deployment events. This may help with this issue as well - https://github.com/change-metrics/monocle/issues/277

Here is from GitHub docs

The value of the **merge_commit_sha** attribute changes depending on the state of the pull request. Before merging a pull request, the merge_commit_sha attribute holds the SHA of the test merge commit. After merging a pull request, the merge_commit_sha attribute changes depending on how you merged the pull request:

- If merged as a [merge commit](https://docs.github.com/articles/about-merge-methods-on-github/), merge_commit_sha represents the SHA of the merge commit.

- If merged via a [squash](https://docs.github.com/articles/about-merge-methods-on-github/#squashing-your-merge-commits), merge_commit_sha represents the SHA of the squashed commit on the base branch.

- If [rebased](https://docs.github.com/articles/about-merge-methods-on-github/#rebasing-and-merging-your-commits), merge_commit_sha represents the commit that the base branch was updated to.

morucci:
Fix #1072

### DIFF
--- a/src/Monocle/Backend/Index.hs
+++ b/src/Monocle/Backend/Index.hs
@@ -88,7 +88,7 @@ instance ToJSON NonIndexedMapping where
       ]
 
 mergedCommitField :: [Pair]
-mergedCommitField = ["merged_commit_sha" .= NonIndexedMapping]
+mergedCommitField = ["merged_commit_sha" .= KWMapping]
 
 data TextAndKWMapping = TextAndKWMapping deriving (Eq, Show)
 
--- a/src/Monocle/Backend/Index.hs
+++ b/src/Monocle/Backend/Index.hs
@@ -61,12 +61,12 @@ instance ToJSON AuthorIndexMapping where
   toJSON AuthorIndexMapping =
     object ["properties" .= AuthorMapping]
 
-cachedAuthorFields :: [Pair]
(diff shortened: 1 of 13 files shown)


src/Monocle/Backend/Index.hs
```diff
@@ -77,6 +77,19 @@ instance ToJSON DateIndexMapping where
...
+
+instance ToJSON NonIndexedMapping where
+  toJSON NonIndexedMapping =
+    object
+      [ "type" .= ("object" :: Text)
+      , "enabled" .= (False :: Bool)
+      ]
+
+mergedCommitField :: [Pair]
+mergedCommitField = ["merged_commit_sha" .= NonIndexedMapping]
```
TristanCacqueray, line 91: Perhaps this should be indexed, so that later we can do filter based on this field?
@leonid-deriv would that works for your use-case?

TristanCacqueray (approved):

src/Monocle/Backend/Index.hs
morucci: Well, I've updated the PR to use the standard keyword mapping for that field.

PR #1076, opened 2023-10-31
Use `committedDate` for ChangeCommitForcePushedEvent's `created_at` field

### LINKED ISSUE #1074: Incorrect commit time in ChangeCommitPushedEvent
both create_at and on_created_at times are times when the PullRequest is created.
Probably on_created_at should be a time when a commit is pushed (obviously this time is available in git and it is visible in GitHub) ... or there should be another field which has a time of a commit.

Any chance this will be fixed? Thank you.

morucci:
Also this change uses `committedDate` for `ChangeCommitPushedEvent`'s `created_at` instead of `pushedDate` which is deprecated.

fix #1074


leonid-deriv:
so both ChangeCommittedPushedEvent and ChangeCommitForcePushedEvent are covered?

TristanCacqueray (approved):

PR #1081, opened 2023-11-05
[release] - 1.10.0

morucci:


morucci:
k8s job fail is expected due to the image not yet published.

PR #1089, opened 2023-11-24
Add author' groups into Monocle Idents

morucci:


src/Monocle/Backend/Test.hs
```diff
@@ -41,11 +41,11 @@ fakeDate = [utctime|2021-05-31 10:00:00|]
 fakeDateAlt = [utctime|2021-06-01 20:00:00|]
 
 alice, bob, eve, fakeAuthor, fakeAuthorAlt :: Author
-alice = Author "alice" "a"
-bob = Author "bob" "b"
-eve = Author "eve" "e"
-fakeAuthor = Author "John" "John"
-fakeAuthorAlt = Author "John Doe/12" "review.opendev.org/John Doe/12"
+alice = Author "alice" "a" mempty
```
TristanCacqueray, line 44: to avoid these changes, we could define a pattern synonym like this:

```haskell
pattern Author :: Text -> AuthorWithGroup
pattern Author name = AuthorWithGroup name mempty
```
… and then rename the data constructor to `AuthorWithGroup`

morucci:
The change is ready to land IMO.

src/Monocle/Config.hs
```diff
@@ -201,6 +204,9 @@ type WorkspaceName = IndexName
 
 type WorkspaceStatus = Map WorkspaceName Status
 
+-- | IdentInfo is the Monocle UID and the list of Groups
+type IdentUG = (Text, [Text])
```
TristanCacqueray, line 208: nit: might be cleaner to use a data record instead of type alias to prevent using the tuple form (like in `getIdentByAliasFromIdents` below)

TristanCacqueray (approved):

PR #1091, opened 2023-11-27
Add support for http proxy

### LINKED ISSUE #1090: Corporate proxy settings for crawler
Hello,

Is there a way to configure crawler to collect metrics from Github, via corporate proxy?

TristanCacqueray:
Fix #1090


TristanCacqueray:
Actually, that may already be the default behavior.

morucci (approved):

PR #1100, opened 2024-01-05
Catch elasticsearch error within the API

TristanCacqueray:
This change improves error reporting when the elastic client throws errors. The implementation is done in the `Monocle.Effects` module.


TristanCacqueray:
Would be nice to have a test (e.g. inject a change without a target_branch before performing a search).

PR #1103, opened 2024-01-06
elastic - bump to latest 7.17.16

morucci:


nix/default.nix
```diff
@@ -138,7 +139,7 @@ in rec {
   # DB Companion
   kibana = pkgsNonFree.kibana7.overrideAttrs (old: rec {
     version = elk7Version;
-    name = "kibana-${version}";
+    name = "kibana-7.17.5";
```
TristanCacqueray, line 142: you need to change the sha256 below, e.g. change a number to get the expected hash

morucci: I don't change the version of kibana into that patch (it raises issue regarding nodejs ... it was the rabbit hole). So the hash is still the same.

TristanCacqueray (approved):

Issue #1112, opened 2024-01-31
Crawler stopped importing data


leonid-deriv:
I have noticed that a crawler stopped importing data. I see the following errors in the log 

```
2024-01-31 14:30:00 INFO    Macroscope.Worker:183: Looking for oldest entity {"index":"demo","crawler":"xxxx-monocle-demo","stream":"TaskDatas","offset":0}
2024-01-31 14:30:00 INFO    Macroscope.Worker:199: Processing {"index":"demo","crawler":"xxxx-monocle-demo","stream":"Changes","entity":{"contents":"xxxx/yyyy","tag":"Project"},"age":"2023-11-29T01:41:16Z"}
2024-01-31 14:30:00 WARNING Lentille.GitHub.RateLimit:66: Repository not found. Will not retry. {"index":"demo","crawler":"xxxx-monocle-demo","stream":"Changes"}
2024-01-31 14:30:00 INFO    Lentille.GraphQL:232: Fetched from current page {"index":"demo","crawler":"xxxx-monocle-demo","stream":"Changes","count":0,"total":0,"pageInfo":{"endCursor":null,"hasNextPage":false,"totalCount":null},"ratelimit":null}
2024-01-31 14:30:00 WARNING Lentille.GraphQL:276: Fetched partial result {"index":"demo","crawler":"xxxx-monocle-demo","stream":"Changes","err":[{"locations":[{"column":7,"line":8}],"message":"Could not resolve to a Repository with the name 'xxxx/yyyy'.","path":["repository"],"type":"NOT_FOUND"}]}
2024-01-31 14:30:00 INFO    Macroscope.Worker:204: Posting documents {"index":"demo","crawler":"xxxx-monocle-demo","stream":"Changes","count":2}
2024-01-31 14:30:00 INFO    Macroscope.Worker:189: Unable to find entity to update {"index":"demo","crawler":"xxxx-monocle-demo","stream":"TaskDatas"}
2024-01-31 14:30:00 INFO    Macroscope.Worker:183: Looking for oldest entity {"index":"demo","crawler":"xxxx-monocle-demo","stream":"Changes","offset":0}
[...]

PR #1129, opened 2024-12-21
Ensure Github PR fetching repo not found non fatal for stream

morucci:
When Monocle discovers organization's repositories it might happen that later repositories are deleted then the Changes crawler may encounter GraphQL queries issues. Such case in now handled and the stream silently continue on that case.

This effort could be subject to followup. But this first change might solve the issue #1112

### DIFF
--- a/src/Lentille/GraphQL.hs
+++ b/src/Lentille/GraphQL.hs
@@ -281,7 +281,7 @@ streamFetch client@GraphClient {..} mkArgs StreamFetchOptParams {..} transformRe
 
         _ <- case dynErrors of
           UnknownErr decodingErrors -> yieldStreamError $ DecodeError decodingErrors
-          NoRepo -> yieldStreamError $ EntityRemoved
+          NoRepo -> yieldStreamError EntityRemoved
           NoErr -> pure ()
 
         -- Yield the results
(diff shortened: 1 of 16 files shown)


src/Lentille/GraphQL.hs
```diff
@@ -44,7 +45,9 @@ import Monocle.Effects
 
 type GraphEffects es = (LoggerEffect :> es, HttpEffect :> es, PrometheusEffect :> es, TimeEffect :> es, Retry :> es, Concurrent :> es, Fail :> es)
 
-type GraphResponse a = (PageInfo, Maybe RateLimit, [Text], a)
+data DynErr = NoRepo | UnknownErr [Text] | NoErr
```
TristanCacqueray, line 48: Perhaps this data could be named `GraphResponseResult`?

src/Macroscope/Worker.hs
```diff
@@ -132,6 +133,7 @@ processStream logFunc postFunc = go (0 :: Word) [] []
       RequestError e -> ("graph", encodeJSON e)
       RateLimitInfoError e -> ("rate-limit-info", encodeJSON e)
       PartialErrors es -> ("partial", encodeJSON es)
+      EntityRemoved -> ("entity-removed", encodeJSON ("" :: Text))
```
TristanCacqueray, line 136: Using "null" might work too:

```suggestion
      EntityRemoved -> ("entity-removed", "null")
```

TristanCacqueray (approved):
Well done!

PR #1137, opened 2025-01-03
Bump to ghc 9.6.6 from Nix latest

morucci:
First tentative to move our Haskell stack to a higher and stable version of the stack. We rely on podenv/hspkgs to provide a nix derivation for a Haskell package set.

Here we attempt a bump to hspkgs for ghc 9.6.4.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -1,6 +1,7 @@
-{ elasticsearch-port ? 19200, nixpkgsPath, hspkgs, self }:
+{ elasticsearch-port ? 19200, nixpkgsPath, latestnixpkgsPath, self }:
 let
   nixpkgsSrc = import nixpkgsPath;
+  latestnixpkgsSrc = import latestnixpkgsPath;
 
   rev = if self ? rev then
     self.rev
@@ -49,7 +50,7 @@ let
   };
 
   # pull latest nixpkgs for just [private] support
-  latestPkgs = import (pkgs.fetchFromGitHub {
+  latestPkgsForJust = import (pkgs.fetchFromGitHub {
     owner = "NixOS";
     repo = "nixpkgs";
     rev = "32ea06b23546a0172ac4e2aa733392e02f57503e";
@@ -58,12 +59,13 @@ let
 
   # create the main package set without options
   pkgs = nixpkgsSrc { system = "x86_64-linux"; };
+  latestPkgs = latestnixpkgsSrc { system = "x86_64-linux"; };
   pkgsNonFree = nixpkgsSrc {
     system = "x86_64-linux";
     config.allowUnfree = true;
   };
   # final haskell set, see: https://github.com/NixOS/nixpkgs/issues/25887
-  hsPkgs = hspkgs.hspkgs.extend haskellExtend;
+  hsPkgs = latestPkgs.haskellPackages.extend haskellExtend;
 
   # manually adds build dependencies for benchmark and codegen that are not managed by cabal2nix
   addExtraDeps = drv:
@@ -359,17 +361,16 @@ in rec {
     [ kibanaStart elasticsearchStart monocleReplStart monocleWebStart ];
 
   # define the base requirements
-  base-req = [ pkgs.bashInteractive hspkgs.coreutils pkgs.gnumake ];
+  base-req = [ pkgs.bashInteractive hsPkgs.coreutils pkgs.gnumake ];
   codegen-req = [ pkgs.protobuf pkgs.ocamlPackages.ocaml-protoc ] ++ base-req;
 
   hs-req = [
     # Here we pull the executable from the global pkgs, not the haskell packages
-    hspkgs.cabal-install
-    hspkgs.fourmolu
-    # Here we pull the proto3-suite executable from the haskell packages, not the global pkgs
+    hsPkgs.cabal-install
+    hsPkgs.fourmolu
     hsPkgs.proto3-suite
-    hspkgs.zlib
-    hspkgs.weeder
+    hsPkgs.zlib
+    hsPkgs.weeder
   ];
 
   # define javascript requirements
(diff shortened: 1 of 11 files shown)


TristanCacqueray:
Hey @morucci , we shouldn't need hspkgs anymore, ghc-9.6 is now well established in nixpkgs. Let's remove that dependency and import the few overrides that are still needed. I think it should mostly work, though we need to fix the paths for the tools like weeder/hlint/fourmolu, which were provided as `hspkgs.$tool` where nixpkkgs provide them as `pkgs.haskellPackages.$tool`.

TristanCacqueray:
I remember doing that, but then a go package failed to build because the infrastructure change a bit. So let's replace the hspkgs import with `github:NixOS/nixpkgs/d3780c92e64472e8f9aa54f7bbb0dd4483b98303` (which is the latest from https://github.com/NixOS/nixpkgs/tree/haskell-updates) to avoid updating the non-haskell things. Later we can fix the rest of the build to only pull a single nixpkgs.

morucci:
Hey @TristanCacqueray thanks for the guidance. I'll try to address that path in followup commits inside that PR. I understand that we'll have a bit of delay using nix to benefit latest Haskell tool chains. So yes if 9.6 and its dependencies are now well integrated in nixpkgs then let's try to use it directly.

TristanCacqueray:
It's not really because of nix, it's actually stackage which govern the set of "stable" versions. I think the default (9.6) should be fine, but you can try a more recent version by replacing `pkgs.haskellPackages` with `pkgs.haskell.ghc910` for example.

TristanCacqueray:
It's odd that the formatting changed, are you using?

```elisp
(use-package nix-mode
  :config
  ;; auto format with nixfmt by default
  (reformatter-define nixfmt-format
    :program "nixfmt")
  (add-hook 'nix-mode-hook 'nixfmt-format-on-save-mode))
```

With `nix profile install nixpkgs#nixfmt`

nix/default.nix
```diff
@@ -49,7 +73,7 @@ let
   };
 
   # pull latest nixpkgs for just [private] support
-  latestPkgs = import (pkgs.fetchFromGitHub {
+  latestPkgsForJust = import (pkgs.fetchFromGitHub {
```
TristanCacqueray, line 76: I think this can be removed, the new "latestPkgs" you just added should contain the right just version, so just removing this binding may be enough.

nix/default.nix
```diff
@@ -358,222 +383,221 @@ in rec {
...
-      export NIX_PATH=nixpkgs=${nixpkgsPath}
-      export MONOCLE_ELASTIC_URL=http://localhost:19200
-    '';
-  };
-  inherit pkgs;
+      shell = hsPkgs.shellFor {
+        packages = p: [ (addExtraDeps p.monocle) p.pretty-simple ];
+
+        buildInputs = [
+          latestPkgsForJust.just
```
TristanCacqueray, line 585:
```suggestion
          latestPkgs.just
```

nix/default.nix
```diff
@@ -1,27 +1,51 @@
...
   haskellExtend = hpFinal: hpPrev: {
     monocle = hpPrev.callCabal2nix "monocle" src { };
 
+    # there is a test failure: resolveGroupController should resolve a direct mount root
+    cgroup-rts-threads = pkgs.haskell.lib.dontCheck
+    (pkgs.haskell.lib.overrideCabal hpPrev.cgroup-rts-threads {
+      broken = false;
+    });
+
+    # Gerrit needs HEAD
```
TristanCacqueray, line 32: that's silly, let's capture a task to make a new release!

morucci:
Looks like the effectful requirement in cabal is not compatible anymore because of effectful being 2.3.1 in the Nix set.

TristanCacqueray:
Should be fine to use according to https://github.com/haskell-effectful/effectful/blob/master/effectful-core/CHANGELOG.md#effectful-core-2310-2024-06-07 and https://github.com/haskell-effectful/effectful/blob/master/effectful/CHANGELOG.md#effectful-2310-2024-06-07

morucci:
> ```emacs-lisp
> (use-package nix-mode
>   :config
>   ;; auto format with nixfmt by default
>   (reformatter-define nixfmt-format
>     :program "nixfmt")
>   (add-hook 'nix-mode-hook 'nixfmt-format-on-save-mode))
> ```

No, I was using the default `nix-mode-format` function. Now using your config, I'm getting the autoformat with nixfmt at save, and looks like it reverted to the initial format.

nix/default.nix
morucci: https://github.com/change-metrics/monocle/issues/1138

morucci:
> Should be fine to use according to https://github.com/haskell-effectful/effectful/blob/master/effectful-core/CHANGELOG.md#effectful-core-2310-2024-06-07 and https://github.com/haskell-effectful/effectful/blob/master/effectful/CHANGELOG.md#effectful-2310-2024-06-07 . Let's bump the bound in `monocle.cabal`

This seems to cause issue with `withEffToIO`. Let's see the CI result.

TristanCacqueray:
Almost there, try to add `SeqUnlift` to the first param of `withEffToIO`

nix/default.nix
```diff
@@ -231,8 +245,8 @@ in rec {
       apiVersion: 1
       datasources:
       - name: Prometheus
-        type: prometheus
```
TristanCacqueray, line 234: That reformat seems wrong, you'll have to manually discard all these chunks.

TristanCacqueray:
One last error!

```
src/CLI.hs:32:41: error:
    Module ‘Options.Applicative.Help.Pretty’ does not export ‘string’
   |
32 | import Options.Applicative.Help.Pretty (string)
   |                                         ^^^^^^
```

nix/default.nix
morucci: Just did a drop of both formating commit. Then re-run the nixfmt (this is the last commit here). That looks fine I guess.

TristanCacqueray:
It looks like the hash function changed, you'll have to update the expected value in the test.

TristanCacqueray:
For fourmolu, here is some elisp for you:

```elisp
(use-package haskell-mode
  :config
  ;; ensure run-haskell uses the simplest ghci subprocess
  (setq-default haskell-process-type 'auto)
  (add-hook 'haskell-mode-hook 'haskell-auto-insert-module-template)
  ;; auto format with fourmolu by default
  (reformatter-define fourmolu-format
    :program "fourmolu"
    :args `("--stdin-input-file" ,buffer-file-name))
  (add-hook 'haskell-mode-hook 'fourmolu-format-on-save-mode))
```

TristanCacqueray (approved):

TristanCacqueray:
well done! Glad to see hspkgs is out of the equation :partying_face:

PR #1148, opened 2025-02-05
crawler - gitlab - make Notes author as nullable

morucci:
We have encountered the following issue when crawling project on a Gitlab instance.

```
"fetch_error":"parse failure: Error in $.data.project.mergeRequests.nodes[0].notes.nodes[0].author: parsing GetProjectMergeRequestsProjectMergeRequestsNodesNotesNodesAuthor failed, expected Object, but encountered Null"
```

This change slightly modifies the graphQL schema for the Gitlab driver to reflect a change in recent graphQL schema where the Notes author attribute can be null.

This change does not update the full schema to avoid potentially more handle and because Morpheus fails to load an the MergeRequestState enum.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - [compose] Bumped Elasticsearch to 7.17.26.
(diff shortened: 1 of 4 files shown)


TristanCacqueray:
Arg that's annoying :/

Issue #1150, opened 2025-02-06
Api/Crawler does not detect config change when config.yaml is a mounted ConfigMap


morucci:
My deployment is running on OpenShift and I spotted that changing the ConfigMap that store the config.yaml file, that the api and crawler pods mount, does not trigger any config reload when the ConfigMap is changed.

```
bash-4.4$ stat /etc/monocle/config.yaml
  File: /etc/monocle/config.yaml -> ..data/config.yaml
  Size: 18        	Blocks: 0          IO Block: 4096   symbolic link
Device: fd00h/64768d	Inode: 1224893829  Links: 1
Access: (0777/lrwxrwxrwx)  Uid: (    0/ monocle)   Gid: (1004760000/ UNKNOWN)
Access: 2025-02-06 10:55:24.772641279 +0000
Modify: 2025-02-06 10:55:24.772641279 +0000
Change: 2025-02-06 10:55:27.304725152 +0000
 Birth: 2025-02-06 10:55:24.772641279 +0000

bash-4.4$ date
Thu Feb  6 11:10:14 UTC 2025

bash-4.4$ grep new-crawler /etc/monocle/config.yaml
bash-4.4$ grep new-crawler /etc/monocle/config.yaml
      - name: new-crawler

bash-4.4$ stat /etc/monocle/config.yaml
  File: /etc/monocle/config.yaml -> ..data/config.yaml
  Size: 18        	Blocks: 0          IO Block: 4096   symbolic link
Device: fd00h/64768d	Inode: 1224893829  Links: 1
Access: (0777/lrwxrwxrwx)  Uid: (    0/ monocle)   Gid: (1004760000/ UNKNOWN)
Access: 2025-02-06 11:19:17.938272678 +0000
Modify: 2025-02-06 10:55:24.772641279 +0000
Change: 2025-02-06 11:19:17.938272678 +0000
Birth: 2025-02-06 10:55:24.772641279 +0000
```

We are using https://hackage.haskell.org/package/directory-1.3.9.0/docs/System-Directory-OsPath.html#v:getModificationTime so I assume that why the reloading mechanism does not trigger.

I would suggest to instead compute a checksum of the config and compare with the loaded checksum instead of the file modification timestamp. Using a checksum will be more robust.

morucci:
The change fixes the issue.

PR #1152, opened 2025-02-06
Config reload mechanism is now based on a checksum computation

morucci:
Instead of file modification date. See #1150.


morucci:
Yes your are right, we can store both change and modification time. I'll revisit that change. Thanks for the review.

TristanCacqueray (approved):

Issue #1154, opened 2025-02-07
Crawler loop stuck since multiple hours (seems related to the GitHub driver)


morucci:
The Macroscope loop is stuck since 12 hours, the last line of log is from the Github crawler. 

```
2025-02-06 17:19:39 INFO    Lentille.GraphQL:235: Fetched from current page {"index":"***","crawler":"***","stream":"Changes","count":25,"total":125,"pageInfo":{"endCursor":"Y3Vyc29yOnYyOpK5MjAyNC0wMy0xM1QxowMM5pWqm6","hasNextPage":true,"totalCount":265},"ratelimit":{"remaining":2922,"resetAt":"2025-02-06T17:22:10Z","used":2078}}  
```

TristanCacqueray:
Perhaps this is related to #1147 , it seems like the managerResponseTimeout changed from ResponseTimeoutDefault ([managerResponseTimeout](https://hackage.haskell.org/package/http-client-0.7.18/docs/src/Network.HTTP.Client.Types.html#managerResponseTimeout)), to https://hackage.haskell.org/package/http-client-0.7.18/docs/src/Network.HTTP.Client.Types.html#ResponseTimeoutMicro .

I wonder if this is working as expected with the openssl manager, and it looks like there might be some issue https://github.com/haskell/wreq/issues/70 . So perhaps we could try using Control.Concurrent.timeout in the doRequest retry logic to force the timeout. Though it would be good to reproduce this issue, and make sure timeout works with the openssl FFI call.

PR #1155, opened 2025-02-07
Use System.Timeout to ensure stalled network request terminates

TristanCacqueray:
Fixes: #1154

### DIFF
--- a/src/Lentille/GraphQL.hs
+++ b/src/Lentille/GraphQL.hs
@@ -41,9 +41,11 @@ import Streaming.Prelude qualified as S
 
 import Effectful.Concurrent.MVar qualified as E
 import Effectful.Retry
+import Effectful.Timeout (Timeout)
+import Effectful.Timeout qualified
 import Monocle.Effects
 
-type GraphEffects es = (LoggerEffect :> es, HttpEffect :> es, PrometheusEffect :> es, TimeEffect :> es, Retry :> es, Concurrent :> es, Fail :> es)
+type GraphEffects es = (LoggerEffect :> es, HttpEffect :> es, PrometheusEffect :> es, TimeEffect :> es, Retry :> es, Concurrent :> es, Fail :> es, Timeout :> es)
 
 data GraphResponseResult = NoRepo | UnknownErr [Text] | NoErr
 
@@ -150,7 +152,11 @@ doRequest client mkArgs retryCheck depthM pageInfoM =
   retryingDynamic policy (const retryCheck) $ \rs -> do
     when (rs.rsIterNumber > 0)
       $ logWarn "Retrying request" ["num" .= rs.rsIterNumber]
-    runFetch rs.rsIterNumber
+    Effectful.Timeout.timeout 75_000_000 (runFetch rs.rsIterNumber) >>= \case
+      Nothing -> do
+        logWarn "Request timeout after 75sec" []
+        error "request timeout"
+      Just x -> pure x
  where
   delay = 1_100_000 -- 1.1 seconds
   policy = constantDelay delay <> limitRetries 7
--- a/src/Lentille/GraphQL.hs
+++ b/src/Lentille/GraphQL.hs
@@ -42,7 +42,6 @@ import Streaming.Prelude qualified as S
 import Effectful.Concurrent.MVar qualified as E
 import Effectful.Retry
 import Effectful.Timeout (Timeout)
-import Effectful.Timeout qualified
 import Monocle.Effects
 
 type GraphEffects es = (LoggerEffect :> es, HttpEffect :> es, PrometheusEffect :> es, TimeEffect :> es, Retry :> es, Concurrent :> es, Fail :> es, Timeout :> es)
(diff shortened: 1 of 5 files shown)


src/Lentille/GraphQL.hs
```diff
@@ -150,7 +152,11 @@ doRequest client mkArgs retryCheck depthM pageInfoM =
   retryingDynamic policy (const retryCheck) $ \rs -> do
     when (rs.rsIterNumber > 0)
       $ logWarn "Retrying request" ["num" .= rs.rsIterNumber]
-    runFetch rs.rsIterNumber
+    Effectful.Timeout.timeout 75_000_000 (runFetch rs.rsIterNumber) >>= \case
```
TristanCacqueray, line 155: This needs to be done L111 instead, where the request really happens.

morucci: See my next commit where I've reworked as you suggested. However as you can see in the output here, it does not seem that the retrying mechanism is working when we hit  the 75 second timeout:

```
time (nix run . -- lentille github-changes --url http://localhost:8080 --project xxx --since 2021-05-01 --human)
Using saved setting for 'extra-substituters = https://change-metrics.cachix.org' from ~/.local/share/nix/trusted-settings.json.
Using saved setting for 'extra-trusted-public-keys = change-metrics.cachix.org-1:dCe8jx9vptiF6DCdZ5y2QouvDsxgFRZnbHowhPnS4C0=' from ~/.local/share/nix/trusted-settings.json.
Couldn't evalue the stream due to The HTTP request hung for more than 75 seconds
CallStack (from HasCallStack):
  error, called at src/Relude/Debug.hs:296:11 in relude-1.2.2.0-1k14DcbZXw3HxDb5R7Nqys:Relude.Debug
  error, called at src/Monocle/Effects.hs:548:16 in monocle-1.11.2-4fvtplU1lRQLlwaeGoiOEa:Monocle.Effects

real	1m16.392s
user	0m1.025s
sys	0m0.422s
```

TristanCacqueray: Good catch, that's because `retryingDynamic` in doRequest or `httpRetry` catch the exception thrown by error. I think we need a new handler in `httpRetry`... when using error, we get the [ErrorCall](https://hackage.haskell.org/package/ghc-internal-9.1201.0/docs/src/GHC.Internal.Exception.html#ErrorCall) exception.

We could use a custom exception for that use case, using something like:

```haskell
data HttpClientTimeout = HttpClientTimeout HTTP.Request
instance Exception HttpClientTimeout
```

Catch it in `httpRetry` with:

```
  timeoutHandler (RetryStatus num _ _) = Handler $ \case
    HttpClientTimeout req -> .. 
```

Now, to throw, we could use `Effectful.Exception.throwIO`. It's not ideal to use exception, and we should prefer an Error effect or (the Except transformer), but since we are already using the exception machinery to handle http error, then we might as well use a custom exception for now.

morucci: Our nix env provides effectul-core 2.3.1:
```
ghc-pkg list | grep -i effectful-core
    effectful-core-2.3.1.0
```

However Effectful.Exception is available in >= 2.4

TristanCacqueray: Arg, ok, well, then let's use `unsafeEff_ $ Control.Exception.throwIO ...`. I think it's ok to use `usafeEff_` from the Monocle.Effects module since that's how we already run other IO action.

TristanCacqueray: Note that's how Effectul.Exception does it: https://hackage.haskell.org/package/effectful-core-2.5.0.0/docs/src/Effectful.Exception.html#throwIO .

Remember, when the IO is on the right side, it's easy to lift the action into Eff, the difficulty is when the IO is on the left side, e.g. `timeout :: IO a -> ...`, that's when we have to use the unliftio technique with `withEffToIO` to adapt the `Eff` into an `IO` so that it can be passed as a callback.

nix/default.nix
```diff
@@ -124,6 +124,21 @@ let
...
+    strict-mutable-base = let
+      src = pkgs.fetchFromGitHub {
+        owner = "arybczak";
+        repo = "strict-mutable";
+        rev = "beea8212f10c9b003e86b208a78de43de8f0442c";
+        sha256 = "sha256-niGBdSrkatr+TZCcLYXo4MDg5FyXTYiKd5K+ZIWSWBs=";
+      };
+    in pkg = (hpPrev.callCabal2nix "strict-mutable-base" src { });
+
+    effectful-core = pkgs.haskell.lib.overrideCabal hpPrev.effectful-core {
```
morucci, line 137: I have trouble getting the right version that expose the Exception module:

```       > Using Parsec parser
       > Configuring effectful-core-2.4.0.0...
       > Error: [Cabal-8010]
       > Encountered missing or private dependencies:
       >     safe-exceptions >=0.1.7.2, strict-mutable-base >=1.1.0.0
       > CallStack (from HasCallStack):
       >   dieWithException, called at libraries/Cabal/Cabal/src/Distribution/Simple/Configure.hs:1457:11 in Cabal-3.12.0.0-inplace:Distribution.Simple.Configure

```

morucci: Ok ... missed your previous comment

src/Lentille/GraphQL.hs
morucci: Yes, I had that idea at some point but that ends up with `Could not deduce ‘Exception e0`

src/Monocle/Effects.hs
```diff
@@ -535,15 +536,31 @@ httpRequest request = do
   HttpEffect manager <- getStaticRep
   unsafeEff_ $ HTTP.httpLbs request manager
 
+httpRequestWithTimeout ::
+  (HasCallStack, Exception e) =>
```
morucci, line 540: Could not deduce ‘Exception e0’

TristanCacqueray: The type variable `e` is unused, try without the constraint.

src/Monocle/Effects.hs
```diff
@@ -565,6 +581,8 @@ httpRetry urlLabel baseAction = Retry.recovering policy [httpHandler] (const act
       promIncrCounter httpFailureCounter label
       pure True
     InvalidUrlException _ _ -> pure False
+  httpHandler2 (RetryStatus _ _ _) = Handler $ \case
+    HttpClientTimeout _ -> pure True
```
TristanCacqueray, line 585: Would be nice to log here and increase the error counter for prometheus.

morucci:
Now that seems to behave as expected :rocket:

TristanCacqueray:
Excellent, well done, exceptions are a bit tricky to get right with Haskell.

